PR #1, opened 2023-03-10
allow linking to a quick start scenario

MischaU8:
Allow linking straight to a scenario without having to use a "scary" `?s=<base64>` url.

For example something that I could use in Discord:
https://lite.koboldai.net/?scenario=KoboldGPT


LostRuins:
LGTM

PR #13, opened 2023-04-14
fix support for correct protocol on https

tony-0tis:
For local requests via https and port 443 window.location.port is empty(''). Also with https the port can be different, not just 443.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2498,10 +2498,13 @@ Current version: 20
 		if (localflag)
 		{
 			localmode = true;
-			let inputport = urlParams.get('port');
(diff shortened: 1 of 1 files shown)


LostRuins:
Hmm, might have to modify the bottom part a bit, https://lite.koboldai.net serves itself over https, but it should be able to connect to a Kobold United Server or KoboldCpp local server via unencrypted http by default too. So I'll have to add some exceptions. otherwise looks good, will test and merge tomorrow.

LostRuins:
Merged with the above modifications

PR #17, opened 2023-05-02
Enables multi-line world-info.

TaleirOfDeynai:
This is a tweak I've been applying to my own `kobold.cpp` copy of Kobold Lite, but it has become onerous due to frequent updates (my merge tool sucks with resolving conflicts in minified source code).

This just allows world-info to be multi-line, which is kind of important for certain kinds of models that benefit from extraneous formatting (like instruct models).  I don't see any reason why it ever should have been an `<input>` instead of a `<textarea>` and it all works and looks fine just switching the elements over, since the JS APIs used by Kobold Lite are congruent between them.


LostRuins:
Is this approach suitable for you?

PR #22, opened 2023-06-18
Idle responses for storymode

neph1:
This adds some things I wished for personally, but also related to this: https://github.com/LostRuins/koboldcpp/issues/228

### DIFF
--- a/index.html
+++ b/index.html
@@ -5193,11 +5193,13 @@ Current version: 41
 	}
 
 	function toggle_opmode() {
-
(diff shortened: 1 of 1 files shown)


LostRuins:
I'll take a look and see if I can add it in without having to duplicate the code

neph1:
Yeah, sorry. Web development isn't really my thing. Let me know if I can improve it.

LostRuins:
Idle responses is added to story mode as a shared panel.

PR #25, opened 2023-08-05
Added optional aesthetic mode for UI of Instruct mode.

Lyrcaxis:
**Changes:**
- 'Aesthetic UI' toggle (previously under 'Chat mode settings') is now on the right panel of the 'Format settings' tab.

**Additions:**
- 'Aesthetic UI' now affects Instruct mode as well.
- Aesthetic mode:
    - Text following the `seq_start` and `seq_end` is wrapped in chat bubbles and styled appropriately.
    - Text outside of chat bubbles (e.g. system messages) is slightly faded out.
    - Markdown mode is supported. If enabled, these settings apply:
        - Expressions (text wrapped in **s) are italicized.
        - Speech (text wrapped in ""s) is italicized and colorized.
        - Code blocks are stylized within black blocks for multiline code or white blocks for singleline.
        - Other text inside bubbles remains white.

**Pictures:**
image
- Markdown disabled:
image
- Markdown enabled:
image

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -2811,7 +2811,7 @@ Current version: 53
 		case_sensitive_wi: false,
 		last_selected_preset: 0,
 		enhanced_chat_ui: true,
-		enhanced_chat_ui_instruct: false,
+		enhanced_instruct_ui: false,
 		multiline_replies: false,
 		allow_continue_chat: false,
 		idle_responses: 0,
@@ -4071,7 +4071,7 @@ Current version: 53
 		}
 		if(temp_scenario.opmode == 4)
 		{
-			if (temp_scenario.enhanced_chat_ui_instruct) { localsettings.enhanced_chat_ui_instruct = temp_scenario.enhanced_chat_ui; }
+			if (temp_scenario.enhanced_instruct_ui) { localsettings.enhanced_instruct_ui = temp_scenario.enhanced_chat_ui; }
 			if (temp_scenario.instruct_starttag) { localsettings.instruct_starttag = temp_scenario.instruct_starttag; }
 			if (temp_scenario.instruct_endtag) { localsettings.instruct_endtag = temp_scenario.instruct_endtag; }
 		}
@@ -5222,7 +5222,7 @@ Current version: 53
 		document.getElementById("tfs_s").value = localsettings.tfs_s;
 		document.getElementById("generate_images").value = localsettings.generate_images;
 		document.getElementById("enhanced_chat_ui").checked = localsettings.enhanced_chat_ui;
-		document.getElementById("enhanced_chat_ui_instruct").checked = localsettings.enhanced_chat_ui_instruct;
+		document.getElementById("enhanced_instruct_ui").checked = localsettings.enhanced_instruct_ui;
 		document.getElementById("multiline_replies").checked = localsettings.multiline_replies;
 		document.getElementById("allow_continue_chat").checked = localsettings.allow_continue_chat;
 		document.getElementById("idle_responses").value = localsettings.idle_responses;
@@ -5381,7 +5381,7 @@ Current version: 53
 		localsettings.trimwhitespace = (document.getElementById("trimwhitespace").checked ? true : false);
 		localsettings.persist_session = (document.getElementById("persist_session").checked ? true : false);
 		localsettings.enhanced_chat_ui = (document.getElementById("enhanced_chat_ui").checked ? true : false);
-		localsettings.enhanced_chat_ui_instruct = (document.getElementById("enhanced_chat_ui_instruct").checked ? true : false);
+		localsettings.enhanced_instruct_ui = (document.getElementById("enhanced_instruct_ui").checked ? true : false);
 		localsettings.multiline_replies = (document.getElementById("multiline_replies").checked ? true : false);
 		localsettings.allow_continue_chat = (document.getElementById("allow_continue_chat").checked ? true : false);
 		localsettings.idle_responses = document.getElementById("idle_responses").value;
@@ -7603,7 +7603,7 @@ Current version: 53
 		}
 
 		// Render onto enhanced chat interface if selected. Currently only applicable to Chat & Instruct modes.
-		let isStyleApplicable = (localsettings.opmode == 3 && localsettings.enhanced_chat_ui) || (localsettings.opmode == 4 && localsettings.enhanced_chat_ui_instruct);
+		let isStyleApplicable = (localsettings.opmode == 3 && localsettings.enhanced_chat_ui) || (localsettings.opmode == 4 && localsettings.enhanced_instruct_ui);
 		let inEditMode = (document.getElementById("allowediting").checked ? true : false);
 		if (localsettings.enhanced_chat_ui && !inEditMode && isStyleApplicable)
 		{
@@ -8709,8 +8709,8 @@ Current version: 53
 									</tr>
 								</table>
 								<div class="settinglabel">
-									<div class="justifyleft settingsmall" title="Switches to an aesthetic messenger style UI.">Aesthetic Chat UI </div>
-									<input type="checkbox" id="enhanced_chat_ui_instruct" style="margin:0px 0 0;">
+									<div class="justifyleft settingsmall" title="Switches to a UI style more suitable for RP. Enabling markdown together with this is recommended.">Aesthetic Instruct UI </div>
+									<input type="checkbox" id="enhanced_instruct_ui" style="margin:0px 0 0;">
 								</div>
 								<div class="justifyleft settingsmall">Enable Markdown <span class="helpicon">?<span
 									class="helptext">Allows the UI to use markdown formatting such as quotes and code blocks.</span></span> </div>
--- a/index.html
+++ b/index.html
@@ -7607,14 +7607,14 @@ Current version: 53
 		let inEditMode = (document.getElementById("allowediting").checked ? true : false);
 		if (localsettings.enhanced_chat_ui && !inEditMode && isStyleApplicable)
 		{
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -7769,6 +7748,33 @@
...
+		// Style *Expressions* and "Speech". TODO: Add customization options
+		newbodystr = newbodystr.replaceAll(/&quot;(\S[^*]+\S)&quot;/g, `<em style='color:rgba(150, 150, 200, 1)'>"$1"</em>`);
+		newbodystr = newbodystr.replaceAll(/\*(\S[^*]+\S)\*/g,"<em style='opacity:0.7'>$1</em>");
+		
+		let youStyle = `</p></div></div></div><div class="incoming_msg"><div class="chat_received_msg"><div class="chat_received_withd_msg"><p>`;
+		let botStyle = `</p></div></div></div><div class="incoming_msg"><div class="chat_received_msg"><div class="chat_received_withd_msg" style='background-color:rgba(20, 20, 40, 0.4)'><p>`;
+		newbodystr = newbodystr.replaceAll(you, youStyle).replaceAll(bot, botStyle);		// Style incoming messages and outcoming messages appropriately.
+		chatbody.innerHTML = newbodystr.replaceAll('\r\n','<br>').replaceAll('\n','<br>');	// Replace the HTML and handle newlines as needed.
+
+		// Show the
```
Lyrcaxis, line 7769:
```suggestion
		// Show the 'AI is typing' message if an answer is pending, and prevent the 'send button' from being clicked again.
```

LostRuins:
will review soon when i have some time

Lyrcaxis:
I intentionally didn't make chat bubbles expand fully, to make later changes easier to review when adding portrait + name.
..but looking back this might be worth doing it now.. not sure when I'll be able to pick up back at it.

index.html
```diff
@@ -4017,26 +4017,11 @@
...
-				localsettings.chatopponent = temp_scenario.chatopponent;
-			}
-			if (temp_scenario.chatname) {
-				localsettings.chatname = temp_scenario.chatname;
-			}
+			localsettings.enhanced_chat_ui = temp_scenario.enhanced_chat_ui;
+			localsettings.multiline_replies = temp_scenario.multiline_replies;
+			
+			if (temp_scenario.chatopponent) { localsettings.chatopponent = temp_scenario.chatopponent; }
+			if (temp_scenario.chatname) { localsettings.chatname = temp_scenario.chatname; }
```
LostRuins, line 4024: @Lyrcaxis 
I understand you are trying to simplify the code, but this is incorrect. The reason it needs to be handled like that is because there is a third state - non-existent value, which can happen for a variety of fields when loading for old format files. Assigning `null` to `localsettings.enhanced_chat_ui` would be a very bad idea, hence the **Strict Equality Check** `===` to eliminate Truthy and Falsy values.

index.html
```diff
@@ -4059,12 +4044,9 @@
...
 		{
-			if (temp_scenario.instruct_starttag) {
-				localsettings.instruct_starttag = temp_scenario.instruct_starttag;
-			}
-			if (temp_scenario.instruct_endtag) {
-				localsettings.instruct_endtag = temp_scenario.instruct_endtag;
-			}
+			localsettings.enhanced_chat_ui = temp_scenario.enhanced_chat_ui;
+			if (temp_scenario.instruct_starttag) { localsettings.instruct_starttag = temp_scenario.instruct_starttag; }
+			if (temp_scenario.instruct_endtag) { localsettings.instruct_endtag = temp_scenario.instruct_endtag; }
```
LostRuins, line 4049: `enhanced_chat_ui` field should not be assumed to exist

index.html
```diff
@@ -8742,6 +8736,10 @@
 							<div class="justifyleft settingsmall" title="Inverts all colors, simple light mode">Inverted Colors </div>
 						   <input type="checkbox" id="invert_colors" style="margin:0px 0 0;">
 						</div>
+						<div class="settinglabel">
+							<div class="justifyleft settingsmall" title="Switches to an aesthetic messenger style UI. Only applicable to Chat Mode & Instruct mode.">Aesthetic Chat UI </div>
+							<input type="checkbox" id="enhanced_chat_ui" style="margin:0px 0 0;">
+						</div>
 					</div>
```
LostRuins, line 8743: Personally I prefer the previous approach of placing this toggle within the dropdown category, because aesthetic mode is inapplicable to adventure and story, and placing the toggle outside implies that it is. People will think it's broken when they toggle it in story mode and nothing changes

index.html
```diff
@@ -7749,24 +7737,24 @@
...
-
-		document.getElementById("chat_msg_send_btn").disabled = document.getElementById("btnsend").disabled;
+	function render_enhanced_chat_instruct(input)
+	{
+		var chatbody = document.getElementById("chat_msg_body");
+		if (!chatbody) { return; }
+		
+		let you = get_instruct_starttag(); let bot = get_instruct_endtag(); // Instruct tags will be used to wrap text in styled bubbles.
+		let youStyle = `</p></div></div></div><div class="incoming_msg" style='width:100%'><div class="chat_received_msg" style='width:100%'><div class="chat_received_withd_msg" style='width:100%'><p>`;
+		let botStyle = `</p></div></div></div><div class="incoming_msg" style='width:100%'><div class="chat_received_msg" style='width:100%'><div class="chat_received_withd_msg" style='width:100%;background-color:rgba(20, 20, 40, 0.4)'><p>`;
```
LostRuins, line 7749: image

I think the colors could be improved, this feels not very readable due to contrast issues.

index.html
```diff
@@ -8679,7 +8662,18 @@
 							   <input type="checkbox" id="adventure_context_mod" style="margin:0px 0 0;">
 							</div>
 							<div id="instructsection" class="settinglabel hidden" style="padding-top: 3px;">
-
+								<!-- TODO: Allow selecting name and picture for yourself and AI. This could be embeded on the left of each message.
+								<table class="settingsmall text-center" style="border-spacing: 4px 2px;	border-collapse: separate;">
+									<tr>
+									  <th>You <a class="color_blueurl" href="#" onclick="selectYourPicture()">🎨</a></th>
+									  <th>AI <a class="color_blueurl" href="#" onclick="selectBotPicture()">🎨</a> <span class="helpicon">?<span class="helptext">The name of the person you want to chat with. Multiple opponents can be specified, creating a group chat, separate their names with ||$||</span></span></th>
+									</tr>
```
LostRuins, line 8670: i'm assuming this is WIP as the function doesn't exist. However if you do plan to add user-uploadable images, i would recommend they be stored as minified base64 strings within the embedded settings on the story JSON - you can refer to the `compressImage` function which I use to pack incoming 512x512 stable horde images into low-res 256x256 base64 encoded data. Ideally this data should not exceed 10kb.

index.html
Lyrcaxis: Text outside `seq_start` or `seq_end` (e.g.: system messages) were left faded out by design, but maybe it's not a great idea to do it like that.. hm..

There's a bug in shown in your picture where "speech" outside chat bubbles doesn't get faded out :D

I'll try something different, maybe wrapping the system message in a 'System' bubble.

index.html
Lyrcaxis: I was thinking they'd just select a path which would be used to load the image each time it was needed.
Let me know your thoughts on this or whether you'd rather them be stored as portable images in the JSON instead!

index.html
Lyrcaxis: Here's how it looks with a proper background and without the fading:
image

Lyrcaxis: Feel free to suggest any specific colors or jump in and adjust them as needed.
I would like to add a full-on color/font customization panel eventually that opens with a click on some 🎨 button maybe, but yeah that's not for now :D

index.html
Lyrcaxis: I think it'd be cool to eventually switch it towards a toggle & (maybe ⚙️) btn that does smth like: 'customize UI from json'.
If/when customization panel is added for the UI, users might wanna store their custom settings and share them with others.
Then 'Kobold-Enhanced-UI' mode could be an option from a dropdown, along with 'Kobold-Classic-UI' mode.

Thought a bit way too much on it without sharing anything and it's embarassing to share xD ..but felt like a cool idea and fairly adjacent to implement along with adding customization options for instruct mode's UI.

Let me know whether you'd like me to move it back and include a separate toggle for it in instruct settings for now or not.

henk717:
Personally I like the default of the old instruct mode since it will be relatable to the people wanting a ChatGPT like experience, while the Chat UI as it was gave that polished whatsapp look. I like the newer options from the PR, but also consider a setting where the aesthetic only applies to the Chat UI (Perhaps a Aesthetic mode dropdown where its off, Chat only, or both?) so its easier to toggle between the two and things like saves with markdown formatting that benefit from the wider display don't suddenly load in the chat UI.

LostRuins:
Yeah I intend to keep the old one as the default for all scenarios as well as the default setting. this aesthetic mode will be something completely new, a standalone UI mode that should not affect any existing scenarios or any other mode/config.

index.html
LostRuins: Hmm maybe let's skip the user avatar loading first, once we have a baseline for this mode, then you can add them in later.

index.html
LostRuins: I'd be concerned about the maintainability of having too many UIs to be honest, but anyway for now I think following the original aesthetic chat UI toggle placement will probably be best, so yeah I think move it back. 

The themes/customizability can be added later outside of the settings panel - it would be maybe a different panel that only appears within the aesthetic mode. 

Ideally I want to keep the aesthetic UI separate from the regular UI (like how it works now, it's parasitic HTML that hooks onto the main ui for input/output) so that any issues with it won't break the main ui.

LostRuins (approved):
LGTM!

PR #29, opened 2023-09-02
Colorpicker improvements + Safari support for macOS + iOS

Lyrcaxis:
### Changelog:
Improved color pickers for Aesthetic Instruct Mode Customization UI, and added support for Safari.
- Colorpickers now pop up directly below the element whose color we're editing.
- Colorpickers now work on Safari for both macOS and iOS.


LostRuins (approved):
lgtm

PR #35, opened 2023-10-08
[WIP] In-Browser Session History

raefu:
This PR adds a dialog that lets you jump to previous sessions or loaded scenarios easily. Sessions are stored in the browser.

Still missing some testing and pruning of older sessions, but the basics all appear to work.

image

Looking for comments.


henk717:
We originally deliberately did not do this, this encourages users to use their browser storage to much which users wont understand is their browser. I expect a lot of unexpected data loss when their browser storage gets cleaned up.

LostRuins:
I am inclined to agree with Henky on this one. I do think users should save to disk rather than browser if possible.

PR #36, opened 2023-10-09
minor cleanups

raefu:
- autosave less
-  reduce console spam
- avoid continuous unnecessary OPTIONS preflights (may slightly improve latency)


LostRuins (approved):
good to go

PR #38, opened 2023-10-11
improve chub scenario preview rendering

raefu:
For some reason, "title" is actually "subtitle", and "name" is the card name. Keeping newlines visible improves readability.


LostRuins (approved):
lgtm will merge

PR #47, opened 2023-12-23
Add `Keep Memory and World Info` option to New Session

one-lithe-rune:
### Motivation

Switching on edit mode, selecting everything, and then deleting, whenever I want to restart story generation from the beginning, or forgetting to do that and losing my just made edits to Memory (or World Info), has gotten tedious enough that I broke down and implemented an option to keep them.

### Changes

* Adds a `Keep Memory and World Info` checkbox box to the new session dialog, defaulting to unchecked.
* Updates `restart_newgame` with an optional `keep_memory` argument, defaulting to false.
* Call `restart_newgame` with `keep_memory` set from the checkbox selection when the New Session dialog is confirmed.

### Possible Problems/Concerns

* Maybe it should be `clear_memory` for the function parameter, and have the not in the call?
* I've set the default to not keep memory, so you get same behaviour as the current version by default. However maybe keeping should be the default? It's what I want to do most of the time.
* Tested connecting locally to Koboldcpp, rather than Horde.

### Screenshot

return_memory_option

### DIFF
--- a/index.html
+++ b/index.html
@@ -7816,15 +7816,7 @@ Current version: 101
 		synchro_pending_stream = "";
 		waiting_for_autosummary = false;
 		last_reply_was_empty = false;
-		if (!keep_memory) current_memory = "";
-		if (!keep_memory) current_anote = "";
-		if (!keep_memory) current_wi = [];
 		pending_context_preinjection = "";
-		if (!keep_memory) extrastopseq = "";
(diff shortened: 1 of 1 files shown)


LostRuins:
There's actually a shortcut for that! Just long press the "Back" button.

Does that satisfy your use case?

LostRuins (approved):

PR #48, opened 2023-12-29
Add option to change the background image

Ar57m:
& adding transparency(maybe too much)

### DIFF
--- a/index.html
+++ b/index.html
@@ -173,7 +173,6 @@ Current version: 102
 			height: 66vh;
 			display: flex;
 			vertical-align: bottom;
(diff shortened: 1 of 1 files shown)


Ar57m:
Nice work👏, today I was trying to do something like this: that after importing the bg image if I close(or refreshes) and open again the page, would be nice the latest bg image to be imported automatically if it wasn't removed in the last session.
I tried a lot using sessionStorage, localStorage and cookies(all with window.onload) but it didn't work(Unfortunately I'm not skilled enough), If you can take a look I would appreciate

PR #57, opened 2024-01-31
XTTS config fixes

Seantourage:
Fixes XTTS setting being overwritten if there are more than 100 TTS voices available on the system.

Stores your xtts_speaker setting between multiple sessions

Setting const default_xtts_base = " http://localhost:8020"; to your endpoint and changing speech_synth: 0, //0 is disabled, 1000 is xtts to 1000 will have the client auto reconnect to XTTS with your prior settings on refresh.


LostRuins (approved):
should be good to go now, please try

PR #58, opened 2024-02-01
Add /slash command system to chat mode, with memory related and /worldinfo commands

one-lithe-rune:
### Motivation

I got to thinking that the model should be able to add and recall things outside its context window for the character it is playing by just saying saying some sort of 'magic words' which lite would parse. After some fiddling about I realised that since this is chat mode, the 'magic words' would probably make most sense as /slash commands.

I thought about reusing world info as the memory store, but that seemed dubious. Instead this implements the memory as a very simplistic dictionary of saved responses, and has a separate /slash command to lookup (but not change) world info 

### Changes

(fdd7da4cbce8b10674d10813f7d9a37069a95dfa)

* Add a basic slash command system.
* Add a /help command for chat mode.

(b321298a9bffe47ea3fcb7f8df9df4a752b1bde3)

* Refactor worldInfo retrieval into its own function
* Add a /memorize command for chat mode
* Add a /recall command for chat mode
* Add a /worldinfo command for chat mode
* Fixes to when and how / commands are parsed for both input and generation.

### To Do

- [ ] localsetting for enabling/disabling chat commands + UI
- [ ] localsetting for name of under which / commands replies are shown (UI?)
- [ ] localsettings for / command response templates (probably no UI atm)

### Possible Problems/Concerns

- /slash commands might not have been the best idea. Although they're probably something any given model probably has the general idea of, because of that they're also something models like to hallucinate.
- Memory system allows memorization of the last response from a specified character, rather than anything arbitrary.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2563,27 +2563,79 @@ Current version: 107
 		}
 	}
 
+	function get_world_info(wimatch_context) {
+		let result = "";
+		if (current_wi.length > 0) {
+			for (var x = 0; x < current_wi.length; ++x) {
+				let wi = current_wi[x];
+
+				//see if this is a valid wi entry
+				if (wi.key == null || wi.key == "") {
+					continue;
+				}
+
+				//selective, but bad secondary key. treat as only 1 key
+				let invalidseckey = (wi.selective && (wi.keysecondary == "" || wi.keysecondary == null));
+
+				let wiks = wi.key.split(",");
+
+				let shoulduse = false;
+				if (wi.constant) {
+					shoulduse = true;
+				}
+				else if (!wi.selective || invalidseckey) {
+					if (localsettings.case_sensitive_wi) {
+						shoulduse = wiks.some(k => wimatch_context.includes(k.trim()));
+					} else {
+						shoulduse = wiks.some(k => wimatch_context.includes(k.trim().toLowerCase()));
+					}
+				} else {
+					let wiks2 = wi.keysecondary.split(",");
+					if (localsettings.case_sensitive_wi) {
+						let t1 = wiks.some(k => wimatch_context.includes(k.trim()));
+						let t2 = wiks2.some(k => wimatch_context.includes(k.trim()));
+						shoulduse = (t1 && t2);
+					} else {
+						let t1 = wiks.some(k => wimatch_context.includes(k.trim().toLowerCase()));
+						let t2 = wiks2.some(k => wimatch_context.includes(k.trim().toLowerCase()));
+						shoulduse = (t1 && t2);
+					}
+
+				}
+
+				if (shoulduse) {
+					result += wi.content + "\n";
+				}
+			}
+		}
+		return result;
+	}
+
 	function slashcommand_help() {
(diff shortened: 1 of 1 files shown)


esolithe:
Like the idea of this - Just a thought from my side though from some quick tests:
- The basic functionality seems to work from my tests:
  - image
  - And gets some basic AI comments to sometimes!
- Some commands, like help, might be better if they prevent actual AI generation I think. As it doesn't make much sense in my opinion to get AI comments on some commands.
  -  image
  - Maybe can be part of the UI as a toggle per command?
- I suspect the recall / memorize would be useful in WI variants to, one of the more common requests I hear often comes down to "Can I save the last response as world info" or highlight a bit and put it into WI. Could be a nice thing to support.
- This is a bit of a longer shot, but I'm also pretty interested in seeing if we could do some sort of hover menu - i.e. starting with a slash opens a menu with command suggestions - though this is way overkill I think so probably one to ignore.

If you think these might be interesting I can probably give them a tinker, I'm pretty intrigued by the ideas to generate world info / potentially even run things like "Summarise the context before into memory" similar to the auto generate memory but with a single line command / clean up.

Just opens up quite a few nice quick options for commands from the keyboard with I'm all in favour 👍 !

one-lithe-rune:
> Some commands, like help, might be better if they prevent actual AI generation I think. As it doesn't make much sense in my opinion to get AI comments on some commands.

Yes, I thought of having the turn stay with whichever of You or the AI entered the command. I think that does make sense. But changing the turn order -- along with getting it to work for Instruct and maybe Adventure -- looked non-obvious enough at an initial glance that I decided to KISS and concentrate on basics. 

> I suspect the recall / memorize would be useful in WI variants to, one of the more common requests I hear often comes down to "Can I save the last response as world info" or highlight a bit and put it into WI. Could be a nice thing to support.

My reasoning on having the memories separate from the world info, was that I didn't think people would want the AI putting random stuff in it willy nilly, and the genesis of the idea was for the AI to be able to do things through the commands. We would probably need to make that an User only command. 

The one world info I did wonder about was whether there needs to be a way to allow you to have world info, so you can get at it with the /worldinfo command, but not have it be inserted in the usual way. But I'm not really familiar with how people use world info.

The couple of extra commands that seemed obvious to me to do at some point would be one to trigger an image generation (so the AI could do it as well as you), and another to grab something from a URL. The summarisation one also makes sense, you would probably need summarisation for the URL one anyway.

I think those are future though. Another thought I had was seeing whether its possible to load up a separate .js file with additional commands, as a kind of command plugin/make your own command system. Not sure whether that is direction we want to go though.

For the current PR, I'm considering whether I should split the enable/disable into separate 'enable for AI' and 'enable for User' options, and whether the other TODOs I have for the local setting stuff should actually be saved with the Memory in the save file rather than as general settings.

one-lithe-rune:
UI Changes, adding to the Basic settings for chat:

lite-pr58-settings-additions

List of /commands:

lite-pr58-slash-command-list

LostRuins:
closed as outdated

PR #59, opened 2024-02-02
Narrate only dialog option

Seantourage:
Adds a new TTS checkbox for selecting whether or not you want the entire response narrated (as before) or only dialog. This option will be enabled by default.

Order of operations:
1) ignore anything between *asterisks*
2) if there are "quotes" in reply it will only narrate those 
3) if there aren't quotes it will narrate everything besides *emotes*

### DIFF
--- a/index.html
+++ b/index.html
@@ -3405,6 +3405,7 @@ Current version: 108
 		beep_on: false,
 		notify_on: false,
 		narrate_both_sides: false,
+		narrate_only_dialog: true,		
 		image_styles: "",
 		image_negprompt: "",
 		grammar:"",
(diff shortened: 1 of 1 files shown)


LostRuins (approved):
will probably refactor it into the speak function

LostRuins:
Btw did you use chatgpt to generate the regex code? seems kind of convoluted.

LostRuins:
I cleaned and simplified the regex.

Seantourage:
Nice catch. Regex is still a bugger for me to figure out so used a number of sources to get it working. Glad it worked out.

PR #61, opened 2024-02-09
Add media query styles for printing

one-lithe-rune:
## Motivation

Printing in story mode still includes a lot the UI, and the main text doesn't seem to flow beyond the first page in non-aesthetic mode (at least for me).

This should fix that for all modes.

## Changes

- Hides top menu when printing.
- Hides input row and chat input holder when printing
- Hides bottom info/status line when printing.
- Hides action menu (if displayed) when printing.
- Allows main text/chat history to flow across multiple pages when printing.

## Possible Problems/Concern

- I've targeted this PR at `dev` branch, I'm not sure whether that is how you want things or it should got to `main`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -265,6 +265,25 @@ Current version: 110
 			resize: vertical;
 		}
 
+		@media print {
+			#inputrow, .cht_inp_hold_outer {
+				display: none;
+			}
+			#lastreq, .lastreq, #lastreq2 {
+				display: none;
+			}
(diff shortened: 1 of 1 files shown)


LostRuins (approved):
LGTM although I would modify it to only apply these styles when printable view is toggled, in case someone actually wants to print the original UI

LostRuins:
Actually I think the current layout is fine, so it's probably good for now unless someone complains.

PR #64, opened 2024-02-18
Fix anote/stopseq line drop

jojorne:
Fixed line dropping of memory menu elements in Android Opera.

Before

After


LostRuins (approved):
LGTM thanks!

PR #67, opened 2024-03-13
Allow hiding Horde privacy disclaimer

Skiwe:
Adds a new checkbox to hide privacy notice at the bottom. It seems to work on my machine, but I can't be certain I've done everything properly since I have no coding knowledge or experience.


LostRuins:
The privacy disclaimer has been relocated to the endpoint selection page. But why would you want to remove it anyway?

PR #68, opened 2024-04-06
Adding Cohere API to the list of API providers

lucasberti:
Cohere has recently launched a new, very good 104B LLM called [Command R+](https://txt.cohere.com/command-r-plus-microsoft-azure/) with 128k context length which is available via their API.

It's possible to use Cohere's models via OpenRouter which is already supported, but you need to buy credits to use them.

At the time of writing this PR, it's possible to use Cohere's API for free indefinitely if you use a Trial API key for non-commercial purposes and stay within its rate-limit.

This PR adds Cohere API as a new option in the API providers dropdown menu, with the possibility to use 3 of their models (Command, Command R and Command R+). It's also possible to define a preamble, which is basically a SYSTEM prompt that is sent every request natively and can be used for jailbreaks.

API reference: https://docs.cohere.com/reference/chat

### DIFF
--- a/index.html
+++ b/index.html
@@ -4947,6 +4947,8 @@ Current version: 127
 			new_save_storyobj.savedsettings.saved_kai_addr = "";
 			new_save_storyobj.savedsettings.saved_kai_key = "";
 			new_save_storyobj.savedsettings.saved_openrouter_key = "";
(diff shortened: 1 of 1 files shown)


LostRuins:
Thanks for this PR. I will review it shortly

LostRuins (approved):
LGTM, thanks

PR #72, opened 2024-05-26
Improved identification of character names.

aaronshenhao:
## Motivation
This aims to address https://github.com/LostRuins/koboldcpp/issues/867, where certain lines of text are misidentified as character names.

## Changes
Updated the regex expressions responsible for identifying character names to make sure that only listed characters (stored in `localsettings.chatopponent`) are matched.

The logic for this is done in 4 places in `index.html`. I renamed the variable names of one of the regex expressions to make them more searchable, but ideally this should be delegated to a function to prevent code duplication and inconsistencies.

## To review before merging
- ~~Line 13353 (used for Messenger UI in Chat mode) does not have a `\n` before the opening bracket, but this seems inconsistent with the other regex expressions, so I added it.~~
  - Apparently, if you add a `\n` in front, like all other cases, it fails to format the character in Messenger UI.
- Does line 13356 ([permalink](https://github.com/LostRuins/lite.koboldai.net/blob/ef21a6f82903221771bfa933c9c752b03f6067d6/index.html#L13356)) need to change as well?

## Potential issues
This will break characters which are not added to the characters list. This would affect users who added characters then deleted them, or for those who manually write custom characters into the story. However, I'm not sure if any users fall into these categories, and it seems like it has always been the intention to only search for added characters.

## Scope of impact
It should only affect Chat mode (`localsettings.opmode == 3`).

### DIFF
--- a/index.html
+++ b/index.html
@@ -10267,7 +10267,8 @@ Current version: 141
 				}
 
 				//examine context to try to determine if there's an existing botname
-				var othernamesregex = new RegExp("\n(?!" + localsettings.chatname + ").+?\: ", "gi");
+				var othernamesregex = new RegExp("\n(" + localsettings.chatopponent.split("||$||").join("|") + ")\: ", "gi");
+
 				var tempfullsearchable = current_memory + current_anote + truncated_context;
 				var foundopponent = tempfullsearchable.match(othernamesregex);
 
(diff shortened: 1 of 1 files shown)


aaronshenhao:
Made it a draft, as I realized that there were two more bugs. The regex expression matches too much. For example, with `Bob: X: Y`, it'll think `Bob: X` is the character name. Also, it fails to recognize characters for the Messaging UI.

henk717:
We have an auto mode and I am one of the users that will sometimes put a third character in. So you'd want to limit this PR to sessions where the character names are defined.

aaronshenhao:
> We have an auto mode and I am one of the users that will sometimes put a third character in. So you'd want to limit this PR to sessions where the character names are defined.

@henk717 Sorry, what's the auto mode? And can you describe how you usually put a third character in?

LostRuins:
You can create a group chat by specifying multiple character names on newlines.

LostRuins:
The issue with this approach is that people (myself included) like to change the chat opponent as the story progresses.

LostRuins:
So i might start the chat with a name, halfway through I wanna change chat opponent and then the story would break.

aaronshenhao:
We could keep track of the names of all characters which appear in the conversation of the save file. Character names are automatically added when they generate their first message.

If a user wants to use an old save file or to paste a conversation, they would have to update the character list manually. This character list should be separate from `localsettings.chatopponent`. If the list does not exist, the old behavior should be used instead. The list is only ever created on the creation of new files.

How does that sound?

PR #79, opened 2024-07-18
let user choose alternate greeting for load_tavern_obj

AndLLA:
if alternate_greetings are available, 
in addition to first_mes,
 let the user choose which message should be used.


LostRuins:
Thanks! This has been a long requested feature.
Will merge first.
I'll probably refactor it a bit to get rid of the async functions in order to support older browsers.

PR #82, opened 2024-07-26
feature: Addition of memory snippets

esolithe:
A variant of the ideas of World Info and Vector based Long Term Memory.  While within Lite we cannot expect vector storage to exist - at least not in its traditional form of embeddings any time soon, it is possible to implement the searching algorithms they use to extract information from the entire game text which are relevant.  These are based on similarity, and after testing a few I settled on Tf-Idf - which is explained more in the links I provided in the comments within the new code.

The general basis looks for frequency of distinctive words, the snippet length and a couple of factors like that - to try and provide a better search while still remaining performant.  While this is still a fairly rudimentary implementation / integration within Lite, it runs entirely browser side and seems to do alright with finding relevant matches.

With the implementation, there is the addition of a new menu under context which allows enabling the new option, modifying the minimum similarity required to find matches and how many matches to include (similar to Top N).  The matches are included with the world info as an authors note formatted block with the different snippets.

While I have tried to generally match formatting / style with the existing code, as this is my first PR for Lite apologies in advance if there's anything out of place!

### DIFF
--- a/index.html
+++ b/index.html
@@ -4423,6 +4423,12 @@ pre code,td,th{padding:0}pre code,table{background-color:transparent}.table,inpu
 		xtc_threshold: 0.2,
 		xtc_probability: 0.0,
 		sampler_order: [6, 0, 1, 3, 4, 2, 5],
+		memsnippet_enabled: false, // Memory snippets off by default
+		memsnippet_numOfSnippets: 3, // Number of snippets to include defaults to 3
+		memsnippet_minSignificance: 0.01, // Must have the bare minimum of significance to be included
+		memsnippet_searchRange: 300, // Tokens to include in search
+		memsnippet_chunkSize: 0, // Size of memory chunks / snippets
+		memsnippet_chunkOverlap: 0 // Overlap between chunks
 	};
 
 	var defaultsettings = JSON.parse(JSON.stringify(localsettings));
@@ -11365,6 +11371,12 @@ pre code,td,th{padding:0}pre code,table{background-color:transparent}.table,inpu
 			logitbiasdict = {};
 			wi_searchdepth = 0;
 			wi_insertlocation = 0;
+			localsettings.memsnippet_enabled = false;
+			localsettings.memsnippet_numOfSnippets = 3;
+			localsettings.memsnippet_minSignificance = 0.01;
+			localsettings.memsnippet_searchRange = 300;
+			localsettings.memsnippet_chunkSize = 0;
+			localsettings.memsnippet_chunkOverlap = 0;
 			current_anotetemplate = "[Author's note: <|>]";
 			regexreplace_data = [];
 			placeholder_tags_data = [];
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -18852,4 +18929,495 @@
...
+		/**
+		 * Max length per chunk
+		 */
+		_chunkSplitSize
+		/**
+		 * Overlap per chunk
+		 */
+		_chunkOverlap
+
+		constructor(amountToTrimForContext, maxLengthOfSearchString = 400, chunkSplitSize = 600, chunkOverlap = 200) {
```
BBC-Esq, line 19001: This can sometimes make a difference if we want to allow a user to change it?

index.html
```diff
@@ -18852,4 +18929,495 @@
...
+		 * Escapes a string to a regex supported format
+		 */
+		_escapeRegExp(stringToEscape) {
+			return stringToEscape.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+		}
+
+		/**
+		 * Gets the last response from the AI and yourself
+		 */
+		getLastResponse() {
```
BBC-Esq, line 19113: Maybe add here to remove all specified characters like punctuation marks?  Not sure if this is the right portion...

```this._lastResponseCache = this._lastResponseCache.replace(/[.,\/#!$%\^&\*;:{}=\-_`~()\[\]]/g, "");```

BBC-Esq:
Hope I did this right...

index.html
esolithe: I do not think this is needed.  Different searchers in future may require those characters - so I would say leave them.  This is negated by the Minisearch lib anyway, as it splits by words and gets rid of special characters anyway as far as I understand it.

index.html
esolithe: Done

LostRuins:
I have yet to try the improved version, which I will review once you two are done with it

BBC-Esq (approved):
Trust your judgment unless there's a specific bug, which I can help troubleshoot.

esolithe:
> Was there something to review or you mentioned that you fixed the issue?

As discussed in the chat on Discord, the requested settings should now be UI options.  Henky is giving it a quick go I believe, and the library used can be found here: https://github.com/lucaong/minisearch 

And the minified build in particular here: https://www.jsdelivr.com/package/npm/minisearch (https://cdn.jsdelivr.net/npm/minisearch@7.1.0/dist/umd/index.min.js)

esolithe:
This should now be ready for another review - hopefully a bit more reliable than last time.  As mentioned on discord, please note the following:

- Seems to take about 10-15 seconds for a 5mb file (which given the max file size of lite at the moment should be fine I think).  There are some areas I saw which might be possible to optimise, but they are mostly at a library level.
- This does not replace / work too well at smaller context sizes - while it does help when you're using 8K and up, due to the amount of text it extracts due to the chunking it comes with some caveats - but it is seems to be a fair bit more stable than it used to be.
- I was going to implement caching of the index, but given it seems to run in 10 second or so for that amount of text, I'll imagine it should be fine for now - maybe a possible future optimisation.

LostRuins:
I have a bunch of changes I wanna make, but I think i'll merge the current version and apply them on top of it so we have something to look back to for reference in future.

PR #86, opened 2024-09-18
fix any name matching in aesthetic theme chat mode

koroal:
### Description
Some inputs get rendered incorrectly under the Aesthetic Theme in the Chat Mode.
The problem manifests itself when the `Chat Match Any Name` option is on, the AI's message contains a line break and the following User's message contains a colon character (:) followed by a space on the first line of the message.
Both the standalone KoboldAI Lite and the one bundled with KoboldCPP currently exhibit this problem.
### Cause of the problem
As I understand it, the problem is caused by the User's name getting replaced in the input first, which causes the check `(?!" + localsettings.chatname + ")` inside of `othernamesregex` no longer to work.
### Proposed solution
I have tried reordering the replacements, so that `othernamesregex` was applied before the replacement of the User's name. The fix seems to be working, and so far I haven't encountered problems.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17304,19 +17304,22 @@ Current version indicated by LITEVER below.
 				othernamesregex = new RegExp("(" + namePattern + "): ", "gi");
 			}
 
-			input = input.replaceAll(mynameregex, '{{userplaceholder}}');
-			input = input.replaceAll(mynameregex2, '{{userplaceholder}}');
-			input = input.replaceAll(mynameregex3, '{{userplaceholder}}');
 			if(as.show_chat_names)
 			{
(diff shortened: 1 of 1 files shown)


LostRuins:
Hi, thanks for reporting this bug. You are right that it is bugged, but the provided solution is not entirely correct - it's bugged because the newline causes the other name regex to match with the entire line that also includes the placeholder. 

I've fixed it, can you check if it works fine for you now?

LostRuins:
If it's not working, please let me know and reopen this issue.

LostRuins:
I haven't seen any other inputs that trigger it. Could you send me an example that's still broken?

LostRuins:
Yes, the problem is the chat format itself is very prone to false positives. 

I've added another patch. Can you try again?

PR #87, opened 2024-09-20
Fixing Mistral Templates

pandora-s-git:
Hi there, fixing the templates as close as possible to the ground truth, you ca find more information regarding the templates in here: https://github.com/mistralai/cookbook/blob/main/concept-deep-dive/tokenization/chat_templates.md

Still needs to be verified, so please dont merge yet!

### DIFF
--- a/index.html
+++ b/index.html
@@ -4556,7 +4556,7 @@ Current version indicated by LITEVER below.
 	{
 		"id":12,
 		"name":"Mistral V1",
-		"user":"</s> [INST]",
+		"user":"</s> [INST] ",
 		"assistant":" [/INST]",
 		"system":"",
 	},
--- a/index.html
+++ b/index.html
@@ -4555,16 +4555,23 @@ Current version indicated by LITEVER below.
 	},
 	{
 		"id":12,
-		"name":"Mistral Gen 1",
-		"user":"\\n[INST] ",
-		"assistant":" [/INST]\\n",
+		"name":"Mistral V1",
+		"user":"</s> [INST] ",
+		"assistant":" [/INST] ",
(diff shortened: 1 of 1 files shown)


LostRuins:
Alright i'm sure you'd know better than me, so I'm merging this once you OK it.

LostRuins:
I think having a duplicate v3 will confuse people. Possible to stick to 1 of them?

pandora-s-git:
I will need to take my time to validate everything, but the main issue is how sensible it is to white spaces, as it is currently it still is not perfect, will see what I can do.

henk717:
We went over back and forth on discord, I think the current PR is whats intended.

pandora-s-git:
actually still an issue just noticed

pandora-s-git:
I think its ready now, sorry for the back and forth, white spacing is delicate, hopefully this should match one on one with the ground truth. If any questions regarding the chat templates you can ping me.

LostRuins:
@pandora-s-git  you didn't answer, what's the reason for having a duplicate v3 and how are they different? People won't know which ones to use.

LostRuins:
Anyone have any objections to these template changes?

pandora-s-git:
@LostRuins Hi! As mentionned, the three of them are slightly different with whitespaces, and those are extremely sensible, its impossible to make it otherwise without them being wrong.

I noticed your comment in https://github.com/LostRuins/koboldcpp/pull/1131#issuecomment-2364929064 , I renammed them because the old template was also wrong, should we still keep it even so?

LostRuins:
Alright then, let's do it your way, we can rename the original file to make it clear.

PR #90, opened 2024-10-09
refresh: Corpo-UI light visual tweaks

Ace-Lite:
image

I tweaked the positions of some of the elements in the corpo style and tried to give more pleasing colouring to various elements. 
I will look at the other styles. I think they could be brought up at the same level of visual coherence while keeping it layout-wise the same. Some of the changes are likely to reflect in other styles, due to shared elements.

*even the user mod example was updated.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2233,6 +2233,9 @@ Current version indicated by LITEVER below.
     	background-position: 8px;
 		background-size: 24px;
 	}
+	.corpo_leftpanel_btn:active {
+		transform: translateY(1px);
+	}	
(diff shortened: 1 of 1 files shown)


Ace-Lite:
Pull request or merge request was made in case of other UI changes by different users.
I would have made it a full package otherwise, but to avoid PR conflicts I decided to do this.

LostRuins (approved):
I'm okay with all other changes, except the top bar gradient, which I think looks a bit odd and i'd probably keep the existing color. Everything else is fine.

PR #91, opened 2024-10-28
UI/UX Corpo: file(image) drop into chat additions, prep for customization and tweaks

Ace-Lite:
Another little batch of progress for the Corpo style. Same reason as before (#90) for adding so little. These changes are exclusive to the corpo style, which would be important to note. Feel free to remove or change anything as before based on preference.

### Features:
**1) Drag & drop the file (images) into the chat window.** 
<-- Kobold Lite seems to support story imports but they are global, which in terms of UX design collides with this a bit (There are no problems, just it is weird that you can drop story JSON outside of the chat window that takes 80% of the screen). However, this could be used for further developments, PDF. / TXT. attachments etc. Just speaking out of my head. -->

**2) Customization dialogue for corpo style.** 
<-- As of right now, it is empty outside of avatars. Plans are minimal, likely I will introduce backgrounds masked with gradient. -->

**3) Aesthetic mode's avatars are a thing in corpo style.**
<-- Just images, you can upload your portrait into corpo customization dialogue -->

**4) If there is no quick save created, the left "Quick Load Save" element is not added.**

### Future plans:
I might focus on some minor QOL features later down the line, I guess some weeks later. I don't have that much time to mess around.

https://github.com/user-attachments/assets/27edfaaf-e729-43e6-a801-35ff1c6e46dd

### DIFF
--- a/index.html
+++ b/index.html
@@ -175,7 +175,7 @@ Current version indicated by LITEVER below.
 		}
 
 		body.connected #topmenu {
-			background-color: #337ab7;
+			background-color: #2876b5;
 		}
 
 		#menuitems {
@@ -2249,6 +2249,12 @@ Current version indicated by LITEVER below.
     	flex-direction: column;
 		padding-left: 2px;
 		padding-right: 2px;
+		transition: box-shadow 1s ease, filter 1s ease;
+	}
+	.corporightpanel[imgdrop-active=true]
+	{
+		box-shadow: 0px 0px 0px 6px #2876b5 inset;
+		filter: brightness(140%);
 	}
 	.corpo_leftpanel_close
 	{
@@ -6354,6 +6360,10 @@ Current version indicated by LITEVER below.
 	//attempt to load a file from disk. could be any format, even images
 	function load_selected_file(selectedFile)
 	{
+		if (document.getElementById("corpodropimgzone").getAttribute("imgdrop-active")) {
+			return;
+		}
+		
 		var selectedFilename = "";
 		if(selectedFile)
 		{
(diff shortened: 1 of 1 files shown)


LostRuins:
Hi, thanks for your PR. First, could we split it into 2 PRs and address them separately.

**PR 1 - Corpo Settings Panel, Avatar Customization, Save Panel Items**
- Settings Panel: Button and Layout looks fine. 
- Quick load save stuff is ok
- Image avatar for user (`human_square`) is currently not set to display correctly.
```<div></div>```

**PR 2 - Corpo image drag and drop **
- Hitbox currently interferes with the importing of character cards. I'd suggest, if we actually want to drag and drop images as an image upload, to only accept them within the chat input box zone (`corpo_chat_outer`). Anywhere else would have the previous behavior of opening the dropped file as a save file instead
- Drag and drop is broken on older browsers, you can't use `[...event.dataTransfer.items].forEach((item, i) => {`, take a look how I did my image upload in `pasteimgcontainer`. In general, we maintain compatibility back to ES6 (Firefox 50).
- Not a huge fan of the slow drag-hover blue border. A bit subjective but i'd either ditch the animation and just use an ultra thin gray border, an overlay like OAI, or nothing at all. But this can be discussed.
image

Misc UI Appearance changes
- Mind explaining why `background-color: #337ab7; ->background-color: #2876b5;`? Elsewhere, `#337ab7` is being used as the default "Kobold Blue"

 (let's focus and complete PR 1 first, since that's less contentious and mostly ok)

henk717:
To me it makes more sense to use a unified drag and drop, where if you drop a KoboldAI compatible save it acts in the classic way while if it can't detect a compatible save it tries to add it as an image.

LostRuins:
That will lead to a very jank experience where importing a bad tavern card ends up loading it as a screenshot instead

Ace-Lite:
Closing then.

Considering it wasn't too much as I am used to requesting bigger pieces of code in "merge requests". I hoped it could be done in a bit bigger burst. Once I have time, I will look into different options for importing and try to discuss this more. Tough, to be honest. 

The whole situation around story/character cards dropping is weird, it superficially makes sense for swapping stories on paper, yet in practice by perspective, you would assume what is happening in chat should take priority over something so global since it is what you see. However, in the case of Kobold, you are likely to swap the stories thanks to its lack of a proper agent selector than to attach something to the message. I was hoping to discuss it, but it doesn't feel like there is an opportunity to talk on Discord in a single channel where it can be swept away and it is not focused on the topic. At least, I sparked the discussion here by creating the space. (Could probably use git issues for this though...)

> PR 1 - Corpo Settings Panel, Avatar Customization, Save Panel Items
> Settings Panel: Button and Layout looks fine.
> Image avatar for user (human_square) is currently not set to display correctly.
`<div></div>`

I wanted to see why was forced the default, but it seems like it just wasn't done. I was hoping to do it differently in future, offering a simple solution now. In case where I would separate it, I would do a little bit more, I guess attempting to move the temp nickname text field there as well?

With the background change, I admit I should have reverted that one before posting, was caught up in everything else. But the logic was to make buttons stand out just a little bit more, considering that some monitors will blend more the contrast, it just makes them less stand out and as a result, it's hitbox readability. Speaking from experience as someone who had to do art on a few IPS and AV panels, the contrast curve can be so wildly different to the point where a darker shade becomes a normal shade, losing the contrast entirely. Of course, I was trying to keep the original intent of it being not distracting intact.

LostRuins:
Alright, are you going to create a new PR for the other stuff then? Should I just cherry pick those changes?

PR #92, opened 2024-11-06
Update index.html - WI context limit dropdown

Pyroserenus:
Adds a simple dropdown for WI context limit as a portion of max context.

I'm at least 70% sure I didn't break anything.


LostRuins:
Hi @Pyroserenus thanks for the PR, however I ended up going with automatic truncation limits when combined with memory to prevent overflow. https://github.com/LostRuins/lite.koboldai.net/commit/13cea1610cefe5c7a1624d25ffdc0988ab1fc75f so it'll use 0.5x, then combine with memory and then again overall limited to 0.8x, i think that's probably a good and hassle free approach.

PR #93, opened 2024-11-19
Update for Mistral V7

pandora-s-git:
Noticed something, by default there is no BOS being prepended? since the templates that are here suppose no BOS is prepended by default, may need to be changed.


LostRuins:
BOS will be silently prepended by the backend (koboldcpp) so it doesn't need to be sent by the frontend. You can verify thing by dumping the tokens with `--debugmode`

LostRuins:
Did you miss something in the PR? It looks identical to the V3 template...

PR #96, opened 2024-12-06
Fix text wrap around avatars in Classic UI mode

Cohee1207:
| Before | After |
| --------- | -------- |
| image | image |

### DIFF
--- a/index.html
+++ b/index.html
@@ -15946,8 +15946,8 @@ Current version indicated by LITEVER below.
 					fulltxt = simpleMarkdown(fulltxt);
 				}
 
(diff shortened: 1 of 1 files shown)


LostRuins (approved):
lgtm, thank you

LostRuins:
image
image

Hmm @Cohee1207  on closer inspection, there seems to be some alignment issues when rendering lists and horizontal rules. Any suggestions?

LostRuins:
Alright, I'll use your approach for the time being then.

LostRuins:
The inline-block broke chat mode too. 
image

I think for now i'll revert back to the original style first, it's a bit too messy to deal with all the edge conditions

PR #99, opened 2024-12-29
fix: Address issues with {{user}} tag not being recognised, along with instruct tags being included in search tokens

esolithe:
Adds support for {{user}} and addresses issues with instruct tags being included as search terms


LostRuins (approved):
thanks

PR #104, opened 2025-02-24
Refactors remote scenario source code for easier maintenance

Desaroll:
The main objective is to consolidate scenario source handling into a smaller area of the code, to allow for easier creation or maintenance.

- The `import_scenario` function now handles the inputBox call, the `scenariodesc` values, and resets temp_scenario to null in case of errors.
- If input fields are correctly filled in on `scenario_sources`, buttons and query arguments can be handled without requiring extra code elsewhere.
  - Added an "aether" query parameter for aetherroom. I hope it wasn't omitted intentionally.
- Changed some `console.log("Error:" + error)` or similar into `console.error(error)`. This logs the stack trace of the error and marks the log message in red.

---

The normal behavior for most users shouldn't have been affected. For errors, some messages changed.

A best effort was made not to break abnormal branches. I tested the normal ones (Success, and bad parameters), but I couldn't find cases to test the code for malformed cards in some platforms. They _should_ keep working, but it might happen that some errors don't get reported correctly to the user.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7799,6 +7799,8 @@ Current version indicated by LITEVER below.
 					}
 				}
 				userInput = userInput.endsWith('/') ? userInput.slice(0, -1) : userInput;
+				userInput = userInput.trim().toLowerCase().replace("https://pygmalion.chat/character/","");
+				userInput = userInput.trim().toLowerCase().replace("https://pygmalion.chat//character/","");
 				return userInput;
 			},
 			fetch: (userInput) => {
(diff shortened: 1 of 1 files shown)


Desaroll:
I can remove the JSDocs comments if needed. I use them to remind me of the expected types for each variable, they don't have any effect on runtime.

LostRuins:
Thank you for doing this - I will review and merge it soon. You are right that some sites have non-standard behavior since they often have some quirks in the way their cards are delivered or formatted.

LostRuins (approved):
lgtm will merge this

PR #113, opened 2025-04-18
Added new option to have context always open and overhaul of character editor to be more helpful

PeterPeet:
This was a change I unfortunately made in one go without separate branches. I truly believe that it improves the character generation and feral handling of Memory and WorldInfo because you can have it open at all times.


PeterPeet:
Henry and I reviewed the change and he found some bugs. So I have to refine that.

PR #114, opened 2025-04-19
Implemented a new beginner friendly Character Creator with optional template

PeterPeet:
Overhaul of the character creator to be more user-friendly in my opinion and give beginners the chance to use a character template. Made this a separate atomic commit.


LostRuins:
Thanks. I tweaked it a bit so the template is an option, but they can still toggle it off to use the original input boxes

LostRuins (approved):

PR #115, opened 2025-04-27
My Ghost Writer integration

trincadev:
## Description

This pull request integrates [My Ghost Writer](https://github.com/trincadev/my_ghost_writer) project adding, or enhancing, text analysis capabilities to/of KoboldAI Lite.

## My Ghost Writer features and changes to KoboldAI Lite

- **[NLTK](https://www.nltk.org/) Integration**: a python backend that uses NLTK (Natural Language Toolkit) to analyse the input text and generate a json array with the words frequency objects
- **[FastAPI](https://fastapi.tiangolo.com/) WebServer**: needed for performing the text analysis
- **Vanilla JS frontend**

preview2
preview1

## Future Work

To further enhance this integration, future development could focus on:

- Search for similar sentences, not only single repeated words
- A thesaurus feature, permitting users to change the duplicated terms
- Enhancing user interface elements

Before any other development however I would like to understand if my project it's a good fit for KoboldAI Lite and, in case, how to integrate it.

I didn't understand what's the correct way to add a flag for show and hide the new UI section (I don't think adding a modal it's a good idea but I'm open to suggestions).


LostRuins:
Hmm. Interesting.

On one hand, I really think this functionality is **quite cool** and would be potentially useful for story writing. On the other hand, I think requiring an entire separate API server running locally alongside lite would greatly limit it's accessibility - most people are just running lite standalone in browser and will not be able to access this feature easily - and lite can also be embedded elsewhere.

As you seem to be doing relatively straightforward word count/n-gram analysis, how about doing it within the page itself in Javascript instead? 
- In my dataset explorer (another project), there is a simple n-gram parser for reference, works for words and phrases of any defined length. Granted it is super basic and better support for other delimiters can be considered.    https://github.com/LostRuins/datasetexplorer/blob/main/index.html#L257
- KoboldAI Lite already embedds lunr.stemmer for stemming capabilities, which can be reused
https://github.com/LostRuins/lite.koboldai.net/blob/de8f9eec364e01b09b174b5fa74e7e1c6310fbfb/index.html#L2402

Also, I'm also not sure why adding a submodule to the repo would be necessary. 

Lastly, the code you've provided is somewhat bloated, though this is a minor issue and can always be refactored later.

trincadev:
Hi @LostRuins, thank you for your response.
I will try to port my backend js. Probably only the PortStemmer you linked me isn't enough, but I think I found an alternative and I will try to use it.

LostRuins:
Sorry if you don't mind me asking - why is it not enough? What operations are unsupported? It's not ideal to have more than one stemming library as it'll weigh down the file. 

Also tagging @esolithe as they might have some interest

trincadev:
Right now I'm on the road so I can't work on the best of my capabilities.

The line you linked me provide us the porterStemmer (that I already use) but I need also a tokenizer. Lucky it's only a regex function that I need to split input text into word tokens and there are plenty of ready ones, e.g.

https://naturalnode.github.io/natural/Tokenizers.html

I only need the time to try and test it. A first JavaScript implementation is here:

https://gist.github.com/trincadev/55d7541fb7ea345ea19c0ace4876772e

I need to change the code embedding the portStemmer implementation you are already using with a tokenizer.

esolithe:
> Right now I'm on the road so I can't work on the best of my capabilities.
> 
> The line you linked me provide us the porterStemmer (that I already use) but I need also a tokenizer. Lucky it's only a regex function that I need to split input text into word tokens and there are plenty of ready ones, e.g.
> 
> https://naturalnode.github.io/natural/Tokenizers.html
> 
> I only need the time to try and test it. A first JavaScript implementation is here:
> 
> https://gist.github.com/trincadev/55d7541fb7ea345ea19c0ace4876772e
> 
> I need to change the code embedding the portStemmer implementation you are already using with a tokenizer.

At the moment the implementation we use is fairly simple:
1. Tokenise by using a simple regex to split on spaces
2. Use the stemmer to reduce the words to the key bit of it (removing -ed and such)
3. Filter by a list of stop words to remove common words which likely do not have much significance from a searching sense (though of course they would affect the semantics of the sentence if using embeddings)

Potentially a similar approach could be used to split the words into tokens for the way you're using - just without the stop words filter probably?

Either way as you say the tokeniser is a pretty rudimentary part to add if needed 👍

trincadev:
Ok, I updated the pull request:

- I now use the existing porterStemmer without any additional dependencies. To achieve this, I derived a new WordPunctTokenizer function from [natural](https://naturalnode.github.io/natural/Tokenizers.html) tokenizers.
- The UI menu within the component I added has been improved to be responsive and works quite well on mobile devices (I tested it only on a 6th generation iPad and with the developer's console at various resolutions).
- The submodule has been removed.

By default, the functionality is deactivated as it occupies one third of the screen.

I have added a settings section under the “Token” tab that includes an option to use an external web server (I will use it instead of the built-in feature because, in my opinion, NLTK provides better results).

LostRuins:
thanks! I'll review soon and merge if no issues!

trincadev:
ok, I changed the layout of the word frequency section. Now I'm separating the list of prefix frequencies from the table of word positions for a better use of space.

LostRuins:
Looking at the most recent code, I notice you are still fetching from `remoteWebServer`? I thought you have removed external dependencies.

trincadev:
Yes, the default function is the embedded one. However, I want to show more clearly what these settings do.
I actually forgot to hide the option for the custom webserver in case the feature on word statistics is turned off (I'll make the change tomorrow in the evening - CET timezone, now I'm on the road again).

trincadev:
> Yes, the default function is the embedded one. However, I want to show more clearly what these settings do. I actually forgot to hide the option for the custom webserver in case the feature on word statistics is turned off (I'll make the change tomorrow in the evening - CET timezone, now I'm on the road again).

As I promised I did these change. The default feature is the embedded one like before. However now, at beginning, I'm showing only the flag needed to enable the word frequency stats div.
1_before_enable

2_mgw_enabled

Also now I display the input text field where to write the custom webserver endpoint only after checking that option: much more clear when users are using the default embedded functionality or the optional custom webserver.

4_enabled_custom_webserver

LostRuins:
I had some time to try this out - however it's not really functioning correctly. 

Here are some of the issues I've noticed on first pass. To reproduce them you can load this save file which contains a snippet from Harry Potter.

1. **Not finding words that are clearly in the text** - A simple search for the term "**Harry**" failed to work at all. Other failed searches include "**before**" and "**muggle**" both of which appear many times in the text.
image
2. **Clicking on a Word Malfunctions** - It seems to want to jump to the position in the text, but this is not working, it's just jumping to a random point in the screen (tested with allow->wallow->swallowed) 
image
3. **Missing words** - Searching "**up**" does not list it as an option within the search even though it appears in the text
image
4. **Does not seem to work for any multi-word phrase** - Even searching something basic like "in the" presents 0 results. In comparison, my own [n-gram viewer](https://github.com/LostRuins/datasetexplorer) can handle this scenario.
image
5. **Destructive to context** - Running the tool instantly destroys all paragraphing in the original text.
image
6. **Inconsistent case sensitivity** - searching any word that has mixed case or upper case fails, however searching lower cased words return upper cased results.
7. **Cramped UI** - The UI is really cramped with multiple windows clipping/overflowing, and it's a struggle to navigate even on desktop (see above screenshots). When used on **mobile devices** it is basically unusable. 
image
8. What is the difference between the Filter and Word Freq buttons? On first glance they are doing the same thing, which is simply updating the search.
9. **Defective stemming** - There is a word "uncle" in the text, however searching "uncle" returns no results. Instead, searching "unc" returns "uncl" as an option among others. 
image

Overall, it's giving buggy results. To test, you may use the json save that I have uploaded here. 
[harry_potter.json](https://github.com/user-attachments/files/20164117/harry_potter.json)

LostRuins:
If you'd like, you can take a look at my n-gram parser function, it's super simple (currently only handles words separated by spaces) but it's pretty easy to adapt it to handle common punctuation as well as common stems. 

n-gram tokenization example. Besides whitespace, adding common punctuation e.g. `.,?!()[]{}:'"%-_~` will cover 99% of cases.
https://github.com/LostRuins/datasetexplorer/blob/main/index.html#L257-L270

simple stemming reference (already in lite, it is 2 lines of code)

https://github.com/LostRuins/lite.koboldai.net/blob/517b84352df45cf4bce0b3a0d24ac355950e9bcf/index.html#L21395-L21396

There is no need to rewrite all the text stemming functions like `textStemming`, the provided library ones are far better.

trincadev:
Ok, I try to respond in order:

1. Doesn't find words that are clearly in the text - Currently the search filter doesn't look for individual words but only stemming (this is one of the future improvements I need to make). Also the included porterStemmer function gives sometimes strange results. Here are your examples with their stemming:

   - “Harry” => harri: 216 repetitions
   - “before” => befor: 20 repetitions
   - “muggle” => muggl: 13 repetitions

2. Clicking on a word does not work - This depends on the length of the rows which could bring the selected words off the edge of the screen. The `setCaret()` method theoretically should scroll to the right position, but it doesn't always work. This is a difficult point to improve by simply moving the position of the scrollbar, I will try to find an alternative solution.

3. Missing words - This is easy, I added a check to filter out words shorter than three characters instead of using an array of stopwords... I will remove it.

4. Doesn't seem to work for any multi-word phrases - Yes, I haven't worked on this at the moment (I will try to study https://github.com/LostRuins/datasetexplorer).

5. Destructive to context - I need to avoid removing double-return carriages, but this initially gave me problems and I haven't checked it again. I can probably just skip the blank lines or something.

6. Inconsistent case sensitivity and lower case - Coat of arms words (the only ones you can look up at the moment) are only lower case, and I haven't thought about users who write in uppercase or mixed case. I will definitely take care of that.

7. Cramped UI - I haven't thought about a phone UI, I need to work on it.

8. The “Word Freq.” button activates the stemming function, while “Filter/Update” filters and sorts the result. It's possible also simply press Enter for filtering/sorting and we could simply activate both the “Word Freq.” and “Sort” choices each time, removing the button "Filter/Update".

9. As before, the input field filters (at the moment) only the list of stemmed words, not duplicate words.

Yes, there are several points to be addressed, but I needed feedback on what is important to get started.
Thank you for your suggestion, I will try to use the provided functions you pointed out. However, I think the most difficult points to handle is the n°2: moving the editor scrollbar at the right point could be difficult.

LostRuins:
Hm perhaps a stemmer is the wrong tool for the job then - as per NLTK this is what they do https://www.nltk.org/howto/stem.html

```
>>> plurals = ['caresses', 'flies', 'dies', 'mules', 'denied',
...            'died', 'agreed', 'owned', 'humbled', 'sized',
...            'meeting', 'stating', 'siezing', 'itemization',
...            'sensational', 'traditional', 'reference', 'colonizer',
...            'plotted']
>>> singles = [stemmer.stem(plural) for plural in plurals]
>>> print(' '.join(singles))
caress fli die mule deni die agre own humbl size meet
state siez item sensat tradit refer colon plot
```

So I think the stemmer is probably working correctly - but the application as a search is either incomplete or unintuitive (why do we need a stemmer anyway, since a substring search should match partial words too?)

trincadev:
Now the points 1, 6, 9 (search) and 3 (missing words) should be ok. The search text fields now works also on actual words, not only on stemming (try looking for "up" or "Harry", case sensitive, for example).

I also removed the "Filter/Update" button (point 8). The button "Word Freq" instead now is hidden unless the custom webserver is active: this way I can still filter the results while requesting updates from the backend only when needed.

I'm still working on the points 2 and 5 (locked by buggy `setCaret()` function with long text rows), 4 (multi word support) and 7 (phone UI).

henk717:
Doesn't have to be in this PR, but feedback from discord was that it would be nice to have a button in your side panel that can submit the words / phrases to the already existing phrase banning field quickly. So they can quickly ban it from continuing to use the overused words.

trincadev:
Yes, that's one of the things I was already thinking.

trincadev:
Ok, with this commit now I fix a bunch of problems:

- fixed wrong results when filtering words using the filter input (now the filter works only on the values - not the keys - of the json output array)
- setCaret(), handle editable mode with nested text within SPAN elements
- scroll to the correct range position when selecting a word with setCaret() also with long text sentences.

Finally I can concentrate myself on a better UI. I have already some ideas and I will post some mockup as soon as possible.

LostRuins:
Hi, it's looking much better, and seems to be coming along nicely! 

I have a small question though - and don't take this the wrong way - but do we actually *need* to use the `remoteWebServer` api call? Seeing as this task seems to be doable perfectly fine in browser, I'm not sure what we really need the remote call for. 

If it's attribution you're looking for, I can still link to your github from within lite for credits, even without the remote server - basically I'm just not sure that the remote call adds much extra value especially since it also creates async waiting complexity as well.

trincadev:
> Hi, it's looking much better, and seems to be coming along nicely!
> 
> I have a small question though - and don't take this the wrong way - but do we actually _need_ to use the `remoteWebServer` api call? Seeing as this task seems to be doable perfectly fine in browser, I'm not sure what we really need the remote call for.
> 
> If it's attribution you're looking for, I can still link to your github from within lite for credits, even without the remote server - basically I'm just not sure that the remote call adds much extra value especially since it also creates async waiting complexity as well.

Ok, thank you. No problem for removing the backend api call. Let me just remove that part of the code and do some testing, there are several references scattered around (e.g. within the settings section).
Also I need to do some more polishing on the mobile interface and support the light theme on corpo UI.

trincadev:
Ok, I finally finished! These commits:

- remove the api call to the "My Ghost Writer" custom backend
- add a mobile menù to save some space on mobile device
- handle responsive devices and custom viewport sizes
- refactor the media queries related to my code to improve the mobile UI
- refactor some of my old code to improve the readability
 
I also did a LOT of tests, both on real mobile devices (iPhone and iPad for now) and with playwright (the tests are within https://github.com/trincadev/my_ghost_writer, feel free to check it out if you want).

Right now the text stats functionality works on:
- Classic UI, of course
- Corpo UI (only editable mode - RAW EDITOR)
- Aesthetic UI (only editable mode)

Sorry about this but I need more time to make it work on Corpo and Aesthetic UI in read-only mode and I preferred a good experience on at least one interface.

Two (partially) unrelated questions:
- there is a reason for which we don't have a simple way to activate a full screen mode? This would help saving more space on mobile devices.
- I would like to add an id and an aria-label on the global mobile menu; this would became like this:

```
                <button title="Main Menu Options" class="navtoggler mainnav" type="button" onclick="toggleTopNav()" id="id-mobile-main-menu-options" aria-label="id-mobile-main-menu-options">
					<span class="navbar-button-bar"></span>
					<span class="navbar-button-bar"></span>
					<span class="navbar-button-bar"></span>
				</button>
```

For some reason, without the id on the "Main Menu Options" button (and I added the aria label for clarity) my mobile playwright tests don't work. It's ok for you? I think this should be irrelevant for anyone that doesn't use playwright.

trincadev:
Mmm. I need some time to test it, I will give you a feedback in a few hours (after work time), I hope.

PR #117, opened 2025-05-08
Fix char spec v2 embedded WI import

Cohee1207:
As per the [v2 spec](https://github.com/malfoyslastname/character-card-spec-v2?tab=readme-ov-file#proposed-fields), `character_book` should be nested into the `data` object.


henk717:
Asked cohee what this fixes, there is a png card we don't import correctly.
Foobar

This PR I can confirm fixes the issue without breaking my test png.

LostRuins (approved):
cohee won

PR #118, opened 2025-05-09
Feature: Randomizing macros {{random:...}}, {{pick:...}}, {{roll:...}}

hu-yijie:
The UI currently offers a simple dice mechanic "Action (Roll)" to inject randomness into the user's query, but there is no way for users to control the output format in the style of [SillyTavern's macros](https://docs.sillytavern.app/usage/core-concepts/macros/#general-macros) such as `pick`, `random`, and `roll`.
This PR adds support for these three macros as a togglable option in the Format Menu (called "Randomizing Tags", but I'm not picky about the name). For compatibility with existing SillyTavern cards that make use of these macros, I've stuck to their syntax and [implementation](https://github.com/SillyTavern/SillyTavern/blob/release/public/scripts/macros.js), though I did take the chance to extend the dice notation slightly.

I had to think about where to replace these macros when they appear in the AI's memory or first message (such as when importing a card). I've put them at the end of `confirm_memory()`, meaning that they get updated along with the rest of the Context data, but there might be a better place for this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3256,7 +3256,7 @@ Current version indicated by LITEVER below.
 		instruct_has_latex: true,
 		placeholder_tags: true,
 		render_special_tags: false,
-		inject_randomness_seed: 1115215124,
+		inject_randomness_seed: -1,
 		request_logprobs: false,
 		persist_session: true,
 		speech_synth: 0, //0 is disabled, 1000 is xtts
@@ -3589,12 +3589,12 @@ Current version indicated by LITEVER below.
 		//truncate to first 3 bytes
 		return hsh.substring(0, 6);
 	};
-	function basic_LCG(seed) { //simple RNG for reproducible dicerolls
+	function basic_lcg(seed) { //simple RNG for reproducible dicerolls
 		var m = 2**35 - 31;
 		var a = 185852;
 		var s = seed % m;
 		return function () {
-			return (s = s * a % m) / m
+			return ((s = s * a % m) / m);
 		}
 	}
 	function import_props_into_object(existingObj, objToImport) { //import new fields from one object into another while preserving exsting
@@ -4328,13 +4328,13 @@ Current version indicated by LITEVER below.
 		text = text.replace(/\{\{\[SYSTEM\]\}\}/g, "").replace(/\{\{\[SYSTEM_END\]\}\}/g, "");
 		return text;
 	}
-	function replace_search_placeholders(text,isstorage=false) {
+	function replace_search_placeholders(text) {
 
 		// Remove any instruct tags as needed to ensure a more accurate search
 		text = remove_all_instruct_tags(text);
 
 		// Replace {{user}} and other placeholders
-		text = replace_placeholders(text,false,false,!isstorage);
+		text = replace_placeholders(text,false,false,false);
 
 		return text;
 	}
@@ -4383,64 +4383,85 @@ Current version indicated by LITEVER below.
 			}
 		}
 
-		if(localsettings.inject_randomness_seed!=-1)
+		if(localsettings.inject_randomness_seed>0)
 		{
 			// define helper functions
 			let dicenotation = function(formula,seed=0){
-				RNG = seed ? basic_LCG(seed) : Math.random;
 				formula = formula.trim();
-				if (/^\d+$/.test(formula)){ formula = "1d${formula}"; }
+				if (/^\d+$/.test(formula)) { formula = `1d${formula}`; }
 				const pieces = formula.match(/^(\d+)?d(\d+)([x*\u00D7]\d*\.?\d+)?([+-]\d*\.?\d+)?$/i);
 				if (!pieces) {
+					return ""; //if the dice test fails we still need to return the original unswapped content!
+				}
+				let RNG = seed ? basic_lcg(seed) : Math.random;
+				let numDice = parseInt(pieces[1]) || 1;
+				let dieSides = parseInt(pieces[2]);
+				if (isNaN(numDice) || isNaN(dieSides) || dieSides < 0 || dieSides > 99999) { //dice above 99999 cause extreme slowdown
+					console.warn("Avoid potential dice issues");
 					return "";
 				}
-				let total = (pieces[1]-0)||1;
-				for (let i=0; i < ((pieces[1]-0)||1); ++i) {
-					total += Math.floor(RNG()*pieces[2]);
+				let total = 0;
+				for (let i = 0; i < numDice; ++i) {
+					total += Math.floor(RNG() * dieSides) + 1; // Standard dice rolls are 1-based
 				}
 				total *= pieces[3] ? parseFloat(pieces[3].substring(1)) : 1;
 				total += pieces[4] ? parseFloat(pieces[4]) : 0;
 				return String(total);
 			}
 			let pickfromlist = function(listString,seed=0){
-				const list = listString.includes('::') ? listString.split('::') : listString.replace(/\\,/g, '{{COMMA}}').split(',').map(item => item.trim().replace(/{{COMMA}}/g, ','));
-				return (list.length===0 ? '' : list[Math.floor((seed ? seed/(2**24) : Math.random())*list.length)]);
+				let RNG = seed ? basic_lcg(seed) : Math.random;
+				const list = listString.includes('::') ? listString.split('::') : listString.replace(/\\,/g, '%%COMMA%%').split(',').map(item => item.trim().replace(/%%COMMA%%/g, ','));
+				return (list.length===0 ? '' : list[Math.floor(RNG()*list.length)]);
 			}
 
-			if(!isgametext)
+			//don't waste time if we dont have any such tags
+			let lowerinputtxt = inputtxt.toLowerCase();
+			if (lowerinputtxt.includes("\{\{pick") || lowerinputtxt.includes("\{\{random") || lowerinputtxt.includes("\{\{roll"))
 			{
-	 			inputtxt = inputtxt.replace(/{{roll\s?:?:?([^}]+)}}/gi, function (m,matchValue) {
-					return dicenotation(matchValue, 0);
-				});
-				inputtxt = inputtxt.replace(/{{random\s?:?:?([^}]+)}}/gi, function (m,listString) {
-					return pickfromlist(listString,0);
-				});
-				stablereg = /{{(?:pick)\s?:?:?([^}]+)}}/gi;
-			}
-			else
-			{
-				stablereg = /{{(?:pick|random|roll)\s?:?:?([^}]+)}}/gi;
-			}
+				let stablereg = "";
+				if(!isgametext)
+				{
+					inputtxt = inputtxt.replace(/\{\{roll\s?:?:?([^\}\n]{1,250})\}\}/gi, function (m,matchValue) {
+						return dicenotation(matchValue, 0);
+					});
+					inputtxt = inputtxt.replace(/\{\{random\s?:?:?([^\}\n]{1,250})\}\}/gi, function (m,listString) {
+						return pickfromlist(listString,0);
+					});
+					stablereg = /\{\{(pick)\s?:?:?([^\}\n]{1,250})\}\}/gi;
+				}
+				else
+				{
+					stablereg = /\{\{(pick|random|roll)\s?:?:?([^\}\n]{1,250})\}\}/gi;
+				}
 
-			let countmap = new Map();
-			let accum = '';
-			let lastIndex = 0;
-			let match;
-			while ((match = stablereg.exec(inputtxt)) !== null) {
-				accum += inputtxt.substring(lastIndex,match.index);
-				const matchtxt = match[1];
-				const count = countmap.get(matchtxt) || 0;
-				countmap.set(matchtxt, count + 1);
-				let hashstr = matchtxt + localsettings.inject_randomness_seed;
-				if (isgametext || (match[0].toLowerCase().substring(2,6) !== "pick")) { hashstr += count; }
-				accum += ((match[0].toLowerCase().substring(2,6) == "roll") ? dicenotation : pickfromlist)(matchtxt,parseInt("0x"+cyrb_hash(hashstr)));
-				lastIndex = stablereg.lastIndex;
+				let countmap = new Map();
+				let parts = [];
+				let lastIndex = 0;
+				let match;
+
+				while ((match = stablereg.exec(inputtxt)) !== null) {
+					const tagType = match[1].toLowerCase();
+					const matchtxt = match[2];
+					parts.push(inputtxt.substring(lastIndex, match.index));
+					let count = countmap.get(matchtxt) || 0;
+					countmap.set(matchtxt, count + 1);
+					let hashstr = matchtxt + localsettings.inject_randomness_seed;
+					if (isgametext || tagType !== "pick") {
+						hashstr += count;
+					}
+					let seed = parseInt(cyrb_hash(hashstr), 16);
+					let replacement = (tagType === "roll" ? dicenotation : pickfromlist)(matchtxt, seed);
+					parts.push(replacement);
+					lastIndex = stablereg.lastIndex;
+				}
+				parts.push(inputtxt.substring(lastIndex));	// Add remaining text after the last match
+				inputtxt = parts.join('');
 			}
-			inputtxt = accum+inputtxt.substring(lastIndex);
 		}
 		return inputtxt;
 	}
 	//if alwaysreplace, then settings are not considered, otherwise checks settings
+	//if isgametext is true, then keeping the random values stable will be prioritized
 	function replace_placeholders(inputtxt, escape=false, alwaysreplace=false, isgametext=true)
 	{
 		inputtxt = replace_instruct_placeholders(inputtxt);
@@ -8038,7 +8059,7 @@ Current version indicated by LITEVER below.
 			gametext_arr = [];
 			if(!localsettings.placeholder_tags) //do a one-time replace instead
 			{
-				greeting = replace_placeholders(greeting, false, true, true);
+				greeting = replace_placeholders(greeting, false, true, false); //this is all onetime replacements, no need for stability
 				combinedmem = replace_placeholders(combinedmem, false, true, false);
 			}
 			if(greeting)
@@ -8098,7 +8119,7 @@ Current version indicated by LITEVER below.
 			let prompttxt = temp_scenario.prompt;
 			if(!localsettings.placeholder_tags) //do a one-time replace instead
 			{
-				prompttxt = replace_placeholders(prompttxt, false, true, true);
+				prompttxt = replace_placeholders(prompttxt, false, true, false);
 			}
 			gametext_arr.push(prompttxt);
 		}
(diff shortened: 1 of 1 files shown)


LostRuins:
Can I ask what's that weird UTF character you added? 
`##�COMMA�##`

hu-yijie:
I copied it directly from SillyTavern's implementation of the macro, it's in the .js that I've linked.
It's just a temporary escape sequence for the regex to split things like "{{pick:Hello!,Aww,What the hell\, man?}}" properly, and could be replaced with any other character that the user isn't going to write.

LostRuins:
alright i'll take a look. generally better to use standard ascii

hu-yijie:
Updated mainly to correct the implementation of the `roll` macro: in SillyTavern, it is not "stable" (like `pick`) but can change between submissions (like `random`).
Since both behaviors can be useful, I've added a custom macro "`toss`" as a stable version of `roll`, completing the quartet.

In this latest implementation, updating the Context will replace the stable macros (`pick` and `toss`) but leaves the unstable macros (`random` and `roll`) alone.
Submitting a generation will replace all macros in the constructed prompt, as well as in the gametext (i.e. what the user sees and edits). But the memory and A/N are unchanged, so instructions that go in there like "Spawn {{roll:1d6+3}} enemies" will continue to work as intended.
Replacement happens *before* the prompt is truncated (unlike placeholder tags), on grounds that `pick` and `random` don't play well with truncation.

Alright, I think the logic makes sense now.

LostRuins:
Alright, we're off to a good start, and thanks for your PR, this looks like a very useful feature!

- Lite already has an existing pipeline for replacing **ST Placeholder** tags (such as **{{user}}** and **{{char}})** so I think we can combine this param into that same pipeline, and then it can be controlled by the same flag that toggles all ST placeholders. 
  - So you can put your `inject_randomness()` inside the existing `replace_noninstruct_placeholders()` instead - it's already used in everywhere that requires placeholder substitutions. 
- I understand the **roll**, **random** and **pick** macros, they seem the same as they appear in ST, but I'm somewhat skeptical to having a new "**toss**" macro. Mostly, its because the existing Character Card ecosystem won't understand or utilize it, so it won't be available outside of Lite and will lack cross-compatibility. 
  - If you really think "toss" is necessary, we can have it, but I would like to keep it compatible if possible, so maybe you can PR "toss" into ST first before we add that here? We can maybe do that in a second PR? 
- But also:  
  - https://github.com/LostRuins/lite.koboldai.net/blob/c5a3b7e7e6229537337e91575feaffb77d7c4548/index.html#L14231 which is  `gametext_arr = gametext_arr.map(txt => inject_randomness(txt,false));` : this is **not** a good idea. Submitting a turn should **never** have automatic destructive effects on the existing story, especially since it can trigger an autosave. 
  - Instead to keep things **stable**, what I suggest we do is generate a new "**inject_randomness_seed**" on session initialization inside the `restart_new_game()` function. 
    - This seed will determine the outcome of **ALL** stable rolls, and since a seeded random **always** gives the same results, this will keep them consistent for the same story - we can keep the placeholder unchanged and don't actually have to modify context. Unstable rolls can simply not use the seed.
  - In fact, we can consider exposing this `inject_randomness_seed` to the user, or possible saving the seed to the savefile as a new localsetting. Think minecraft world seeds.
- Lastly, could consider some small touch ups for the regex parsing for the macros - I can look into that later, nothing major.

hu-yijie:
Thanks for the feedback! I like your idea of using a seed to fix every stable roll (especially compared to the alternative of mutating the existing story), but I'm not sure if this can be reliably enforced when the number and order of the rolls can change.

For example, say the story begins with "Bot: I have {{pick:2,3,4}} cats" (which the PRG fixes to 3) and the user edits it to "Bot: I have {{pick:2,3}} dogs and {{pick:2,3,4}} cats". Unless there's a clever implementation I'm missing, this is going to change the cat roll, and I'm not sure if this is acceptable to the user. Especially if they've already continued the chat about those three cats.

But assuming this can be made to work, I'm in agreement with your points:
- The `inject_randomness()` function can be folded into `replace_noninstruct_placeholders()`, no need for a separate function
- The seed can be stored as `localsettings.randomness_seed` and exposed by Lite in the Placeholder Tags section
- No need for a "toss" macro at this stage (I realized that what I really wanted with it was to have controllable re-rolls, and I can do that just by changing the exposed seed)
- OK if you want to clean up the regex parsing (I'd like to keep my dice notation extension `NdX*Y+Z` with float Y and Z, though)

I may not be able to work on this immediately (and perhaps not until I've figured out the roll ordering problem to my satisfaction), so feel free to make your changes first.

LostRuins:
So, if we look at the docs you linked for ST above:
- `Alternative to random, but the selected argument is stable on subsequent evaluations in the current chat if the source string remains unchanged.`

I think this also implies the same - rolls are stable until text is changed, then all rolls after that point of modification can change unpredictably.

**However, I like your idea**. Hashing the roll itself should give great results. Though, when we `pad it before hashing` it's still potentially subject to the same issue as a basic seed, although to a much lesser degree. That would mean that a roll is stable for that particular sequence of words, but if a new instance of that word was appended early in context it would "push down" all other instances of that word. I think that is probably still a good enough approach.

So to summarize, example prompt:

> I have {{pick:2,3,4}} number of {{pick:yellow,green}} cats in my {{pick:car,bus,truck}} which has {{pick:2,3,4}} wheels and is painted {{pick:red,yellow,green}} in color. The traffic light is now {{pick:red,yellow,green}} so it's time to go.

After post processing, assume our **inject_randomness_seed** is 888, it will evaluate to approximately (just for illustration):

> I have `rand_from_seed("s888_n1_pick:2,3,4")` number of `rand_from_seed("s888_n1_yellow,green")` cats in my `rand_from_seed("s888_n1_pick:car,bus,truck")` which has `rand_from_seed("s888_n2_pick:2,3,4")` wheels and is painted `rand_from_seed("s888_n1_pick:red,yellow,green")` in color. The traffic light is now `rand_from_seed("s888_n2_pick:red,yellow,green")` so it's time to go.

the values for `s` come from the story seed, and `n` is determined at runtime through a simple counter within `inject_randomness` which increments a set/hashmap of strings for lookup.

hu-yijie:
ST's distinction between stable and unstable rolls seems to be made more for card creators. From the user's perspective, ST gives each turn its own source string which can be independently edited, with all rolls evaluated immediately once the edit popup is closed. (This also limits the potential damage of instability.) So the distinction is invisible to users: not only are all rolls "stable" from their POV, they can't even be re-rolled without rewriting the tag!

Thinking about this a bit more, I can't imagine `roll` being unstable if it's going to be used for RPG simulation. If I, as User, type "Suddenly, {{roll:1d3}} cats appear" and let the AI respond based on that, it has to be stable for the rest of the session. (On the other hand, when the Author's Note instructs the AI to "Introduce {{roll:1d3}} cats in every reply", we want *that* to be unstable.)

So it seems like we'll need to emulate the ST distinction between context and gamestate when evaluating placeholders. Fortunately, I *think* we can do this by extending `replace_placeholders()` with a boolean flag, which gets passed down to `replace_noninstruct_placeholders()`. This flag would activate when replacing gametext but not memory or A/N, forcing `{{roll}}` to be treated as stable.
I'm not sure if it should force `{{random}}` to be stable as well. There's no precedent from ST since macros are immediately evaluated, so it really comes down to the Lite user's expectations when they use `{{random}}` in their story. How 'random' do they want to be?

LostRuins:
>Edit to remark: Since {{roll}} is more affected by the repetition issue than {{random}}, perhaps we could take a small window of characters around the literal before hashing. Kind of like an inpainting boundary. 

Tbh that's probably overkill. I think it's easier to just keep rolls inherently stable as we describe above. Instead, if a user wants some "dynamic" aspects to random events, they can simply create multiple world info entries with different probabilities of being injected. I think that keeps things simple and doesn't risk messing up past context at all.

Remember that the char card itself may be stable for the session, but the moment the user clicks "new game" their randomness seed will change anyway, resulting in a completely new and dynamic session - not every play session will end the same way.

henk717:
No the whole point of stuff like this is to spice up world info stuff easily. I would not want it to be static in world info that defeats the point of having random rolls there. Having a way to do that is a must.

LostRuins:
It is random. it's a stable random.

LostRuins:
Also you can use the cyrb_hash for any hashing needs you have. It's already built into lite and used for all hashing within

henk717:
> that means that during subsequent turns when you resubmit the same prompt your initial roll **does not change**. Having it change will be extremely disruptive to the coherence of the story.
> 
> If ` {{pick:troll,ogre}}` generated troll on turn 1, and ogre on turn 2, then it should still generate the same things in the same order on turn 3. otherwise your story will keep changing constantly every time you submit it.

Sometimes we want it to change in a world info scenario such as randomizing drop chance for example.

LostRuins:
yes it does, see above for how the hashing works.

hu-yijie:
Okay, I think we now have a working implementation. Whether this is the fully desired behavior, I'm not so sure.

I added back a PRG after all, because I see  no easier way to roll a user-specified number of dice from a single seed. I just used the LCG in the SO thread linked earlier.

Playing around with this feature, I found a strong urge to reroll by pressing "Retry", despite knowing that the randomness and sampler seeds are unlinked. Sadly I can't think of any way to make this a workable default, short of actually evaluating and permanently replacing the macro (which might be acceptable if it's in `submit_generation()`, but then the point of `inject_randomness_seed` as reproducing all randomness is lost), and I'm not sure what users would prefer either.

Having a seed that reproduces all randomness also affects the semantics of stable/unstable rolls. In ST, all placeholder tags are actually evaluated (like in my original gametext-mutating implementation), so it makes sense to have a different kind of randomizing tag that evaluates the same way each time it's inserted into **the same chat session**.  This is not what you want for RPG dicerolls, but it makes sense for things like mystery stories where AI picks a culprit and hides it from User, so card creators can write scenarios like
>Each night, {{pick:Elara,Seraphina,Niko}} will attempt to murder someone, for example by summoning {{roll:1d3+1}} {{random:troll,ogre,cat}}s to devour them.

I've implemented this by removing the counter when `{{pick}}` is inserted from outside the story.

Implementing `inject_randomness_seed` effectively makes everything stable by default, relying on the counter (i.e. number of times a roll appears previously) to inject variation/"instability" in rerolls. In particular, if `{{random:troll,ogre}}` occurs in a World Info entry (with my implementation, you probably don't want `{{pick:troll,ogre}}`) and this entry is always inserted at the top of every submitted prompt, then it'll always pick the same "turn 1" choice, and the only way to change this is by using a different randomness seed.
(At least, I think it does. I'm not completely familiar with all the places in which `replace_noninstruct_placeholders()` is being called, so please let me know if there are any mistakes in where I've inserted the boolean flags.)

LostRuins:
> in rerolls. In particular, if {{random:troll,ogre}} occurs in a World Info entry (with my implementation, you probably don't want {{pick:troll,ogre}}) and this entry is always inserted at the top of every submitted prompt, then it'll always pick the same "turn 1" choice, and the only way to change this is by using a different randomness seed.

Hmm, I'm not sure why that is needed. If the entry is always inserted at the top of the submitted prompt, it would be stable *anyway*, since the relative positions are the same. I.e. If WI is inserted as turn 3, it's always turn 3 and will be stable for the current story. The only time I can see this breaking is if the WI insert location changes over time, until it slips past another instance of the same identifier, which seems extremely unlikely.

**Nonetheless, i'll leave the choice up to you**. My opinion is that the original approach works fine and keeps it simple.

Regarding `submit_generation`, I guess I am ok if you wish to do some *permanent* replacements for submitted text inside `submit_generation`, however, the any past **context** for historical turns cannot be changed. If you want this, you can apply submitted text mutations to `newgen` (the newest submitted text) after https://github.com/LostRuins/lite.koboldai.net/blob/main/index.html#L14375-L14387 inside `submit_generation`. Note that this only applies to *user newly submitted text*, and I don't think users will handwrite `{{macros}}` very much themselves, so this is probably of very limited use...

One last place we have not considered is the **Card Importer** function itself. For that, please take a look at the `load_tavern_obj()` function which handles the *one-time Irreversible Importing** from a png/webp character card. Note that this is an *import* not a load, aka its intended to be an irreversible process. If you want, you can apply any **permanent text transformations** there as well. But they will necessarily only apply to tavern image cards, and will obviously become fixed after import.

henk717:
If WI is inserted in turn 3, but turn 2 is conditional wohld it not become turn 2?

hu-yijie:
>I don't think users will handwrite {{macros}} very much themselves

Maybe not `{{pick}}` or `{{random}}`, but my original motivation for submitting this PR was that I wanted to use ST's `{{roll}}` to customize my "Action (Roll)" submissions...

In any case, I think this discussion is getting more complicated than it has to be. Let me try and summarize:

- As I see it, a text with randomness can be either **gametext** (aka "context", "story", "game") which is a list of turns that the user can see and edit, or **non-gametext** (a catch-all for memory, author's note, world info/lorebooks, websearch, etc.) which is not part of the gametext itself, but gets injected into the gametext to form the final submitted prompt.
- **Randomness in gametext must always be stable.** This can be enforced either by evaluating all macros immediately (which is what ST does), or leaving it unevaluated but **using a seed to enforce consistency** (which is what we're going with). With a seed, replacement becomes an immutable operation, and can be implemented completely in `replace_noninstruct_placeholders()` with **no need for permanent replacements**.
- **Randomness in injected non-gametext can be either stable or unstable.** Sometimes we want consistency (mystery stories), sometimes we don't (RPG dicerolls). We follow ST in naming the macros `{{pick}}` and `{{random}}` respectively.
- The diceroll macro `{{roll}}` **should be stable when it occurs in gametexts, and unstable when it occurs in injected non-gametexts.**

My earlier implementation of `submit_generation()` had a bug which caused non-gametext to be imported as gametext (i.e. no unstable randomness), this has been fixed now. If you start a new session (empty gametext) and put
>A wild {{random:troll,ogre}} appears!

in any non-gametext field (memory, A/N etc.) and submit
>I brandish my {{random:sword,staff}}!

Then the AI will generate a different enemy (which you can reroll just by pressing "Retry"). But your weapon will always be the same (unless you change the seed), even if you retype and resubmit the macro.
But changing the non-gametext to
>A wild {{pick:troll,ogre}} appears!

will keep the enemy consistent as well. (Again, the weapon roll is part of gametext, so it never changes.)

I think this should settle it then?

LostRuins:
Yeah, I should be good with the current PR for now, we can refine again later.

> Randomness in injected non-gametext can be either stable or unstable

mostly correct, but do remember that some injected non-gametext can appear permanent (e.g. WI that is static or always matching and injected) you might want it to be stable too. I would go as far to say that Memory *should always* be considered gametext since its ALWAYS at the start. Memory should thus always be stable.

hu-yijie:
Just a thought that came to me: if the randomness seed is going to be saved in `localsettings` so that users can reproduce their rolls, we should probably save the version number `LITVER` along with it (I don't see it in there at the moment).

This would let us refine the mechanism in future versions without breaking the seed for existing versions' savefiles.

LostRuins:
Not really needed, it won't be helpful to match against imho. 
If future enhancements allow for refinement, we can always add additional data then, and treat earlier saves without this data as legacy.

LostRuins (approved):
- Added some corrections to what is/is not considered gametext, memory was incorrectly labelled as non-gametext, while some onetime substitutions were labelled as gametext
- Added a max length of 250 chars when handling formulas (this helps mitigate the risk that one bad formula e.g. incorrectly terminated {{roll: without a closing }} will break all subsequent text), and also fixing some incorrect regex.
- Search placeholders are always treated as non-gametext, this is unimportant as search queries are sent to a seardh engine as a tiny standalone text snippet, so will never be able to be stable.
- A StringBuilder concat pattern is used to construct the replaced regex, otherwise it becomes extremely slow on large strings with `str += str2` append pattern (in JS it reallocates the whole string every time!)
- Set inject_randomness_seed to -1 by default, as an experimental feature I think we can leave it off by default first until we know it's stable.

Noticed some flaws with our current approach:
 - Since the noninstruct placeholder replacement works differently (called in different sequence different number of times) between corpo mode, aesthetic mode and classic mode, the submitted gametext can potentially be different depending on which UI is selected (or if say its used for a summary). We could consider exposing `countmap` as an optional input so that multiple corpo turns can reuse the same countmap which fixes this issue.
 - Especially with a multi-turn replacement done like in corpo mode, because the Map is countmap is generated within the convo block, every convo block starts with the same countmap which means the replacements will follow the same order as well.
 - We can consider using a different seed offset for memory to prevent it from following the same pattern as the main gametext. 
 
I realized I've probably dragged this along for too long already, so I think i'll go ahead with merging what we have first.
Please take one more look, and if everything is good I will merge it in a day.

LostRuins (approved):

hu-yijie:
This looks good, thanks! My only concern with this version is that the
>max length of 250 chars when handling formulas (this helps mitigate the risk that one bad formula e.g. incorrectly terminated {{roll: without a closing }} will break all subsequent text)

may be too easily breached by existing ST cards, since the common practice for card creators is to repeat `{{random}}`/`{{pick}}` items to skew their selection probabilities. (For example, [this battle simulator](https://www.chub.ai/characters/qwerty213/battle-simulator-d2ed092e) is ~600 tokens long.)

I'd prefer to just raise the limit to something like 2000 chars: it'll be more disruptive when you `break all subsequent text`, but the immediate feedback also means you're made more aware of your error, and I think this is better for users than throwing a parsing error on an actually valid regex.

(This assumes that the bad formula is in gametext. I expect injected non-gametext to be already limited in size, which mitigates the downsides of a higher char limit.)

---
>Since the noninstruct placeholder replacement works differently (called in different sequence different number of times) between corpo mode, aesthetic mode and classic mode, the submitted gametext can potentially be different depending on which UI is selected (or if say its used for a summary). We could consider exposing `countmap` as an optional input so that multiple corpo turns can reuse the same countmap which fixes this issue.

That makes sense, `countmap` feels like it belongs in the same scope as the gametext, which suggests that maybe it should be a global var. Presumably injected non-gametext would continue to have their own countmap (not sure if this would addresses your last point on treating the memory differently? If following the same pattern as the main gametext is a concern, something like `let hashstr = matchtxt + (localsettings.inject_randomness_seed+isgametext)` should break it up).


But yeah, I agree that it'd probably be easiest to just release this as an experimental feature first, and gather feedback on how to take it from here.

LostRuins (approved):

LostRuins:
So @hu-yijie now that this has been merged for a few days, how is it working for you? Any issues?

hu-yijie:
Well, my typical use of it has been pretty basic: mostly `{{roll}}` as a replacement for "Adventure (Roll)" where I can skew the distribution depending on my current situation. (When I don't need this ability, then Dice Mode still works well enough and requires less typing, so it's still worth keeping around.)


I'm skewing with just a simple additive offset like `(Roll: {{roll:d100+50}}/100)`. Luckily my AI can roll with things like `140/100` or `-83/100`, so this is good enough for me.
(In case this issue ever resurfaces, the more principled solution would be to implement [proper dice notation](https://en.wikipedia.org/wiki/Dice_notation#Standard_notation) with "H" and "L".)

Every now and then I'll reroll a submission just for curiosity, which I can do with the floating-point hack (`{{roll:d20}}` -> `{{roll:d20*1.0}}` -> `{{roll:d20*1.00}}` etc.).
My original motivation for wanting floats was to simulate a financial trading game (roll for market movements, LLM to generate analyst reports), although I've since learnt that [you really are giving up fixed-point precision, in Javascript at least](https://stackoverflow.com/questions/1458633/how-can-i-deal-with-floating-point-number-precision-in-javascript). Luckily (again) my AI can go along with values like `0.29000000000000004%`, and I don't think it's worth fiddling with the output just for this one usecase.


I haven't really tested the other macros outside of loading a few Tavern cards, but it seems like the implementation works as intended (other than on overly-long lists). I never use Corpo mode myself, so I'm trusting that you've got that part worked out.
I think this is one of those features where the usecases are emergent rather than known in advance, so probing the edgecases is better left to someone more imaginative than myself. That's why I'm for the idea of releasing this as-is first, and letting other users play around with it, so that they can suggest the best tweaks/extensions/PRs (depending on the suggestion, I may also chip in).

PR #121, opened 2025-05-24
feat: Enhancements to TextDB, WI and embedding support

esolithe:
## What's new
- Addition of document uploading to TextDB for certain types (text files, lorebooks)
- Splitting of text DB into documents
- Importing of lorebooks from the load button into text DB
- Download / Upload of WI groups
- Addition of embedding support
- Overhaul of the UI in the TextDB tab to make it more user friendly

Please note this requires a small KCPP change:
```python
    return {"result":"KoboldCpp", "version":KcppVersion, "protected":has_password, "llm":has_llm, "txt2img":has_txt2img,"vision":has_vision,"transcribe":has_whisper,"multiplayer":has_multiplayer,"websearch":has_search,"tts":has_tts, "embeddings":has_embeddings, "savedata":(savedata_obj is not None), "admin": admin_type, "guidance": has_guidance}
```
Last line of get_capabilities needs to be changed to include if embeddings are enabled.

Please note this does not have compatibility with Firefox 50 which is one of the lite requirements - specifically, the use of async / await for indexDB calls, along with classes (they have been kept modular at the end to allow for easy movement / insertion as utils).

This PR also does not include a search with nearby text.  I debated it, but am unsure if we would want it (given the vectors can be more costly to compare and generally give better results anyway).

### DIFF
--- a/index.html
+++ b/index.html
@@ -13144,7 +13144,7 @@ Current version indicated by LITEVER below.
 		let currentChunkSize = Number(document.getElementById("documentdb_chunksize").value);
 		let snippetEstimate = currentChunkSize * 1.5 * Number(document.getElementById("documentdb_numresults").value);
 		let numberOfChunks = Math.ceil(document.getElementById("documentdb_data").value.length / (currentChunkSize - (Math.min(currentChunkSize * 0.5, 500))));
-		document.getElementById("documentdb_snippetestimate").innerText = `Estimate of context usage: ${snippetEstimate / 3} tokens (${snippetEstimate} characters). Total chunks: ${numberOfChunks}.`;
+		document.getElementById("documentdb_snippetestimate").innerText = `Estimate of context usage: ${snippetEstimate / 3} tokens (${snippetEstimate} characters). Total chunks: ${numberOfChunks}. Cached chunks: ${Object.keys(embeddings_cache).length}`;
 	}
 
 	function confirm_memory() {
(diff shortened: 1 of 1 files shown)


LostRuins:
Do let me know - in case you might not want me making changes to your main branch, you might want to change to a different branch

esolithe:
> > Last line of get_capabilities needs to be changed to include if embeddings are enabled.
> 
> This should already be in the latest kcpp unless im missing something.
> 
> I will add compatibility with firefox 50. Do you want me to push fixes to your main branch (currently your PR is on your main branch) or do you want to use a different branch for it?

Ah nice, in that case feel free to ignore that kcpp change - maybe I got mixed up between the capabilities I added specific to Esobold 😄 

And thanks in advance for the Firefox 50 fixes, feel free to push to the main branch - my lite fork is just for PRs, so all the changes are welcome!

LostRuins:
What about unused functions, should I keep them? I see there seem to be references to `/api/extra/extractText` but I don't see that actually being used.

esolithe:
> What about unused functions, should I keep them? I see there seem to be references to `/api/extra/extractText` but I don't see that actually being used.

It's for the extraction of text from other file types - might be a separate PR - apologies I probably missed removing that so please do 👍

LostRuins:
Alright @esolithe sorry for the delay, this took longer than expected! Your enhancements are now working correctly in firefox 50, so that's nice - I did make quite extensive modifications, but I did so mostly piecemeal, so you can pick and choose the ones that you want to keep for your fork. 

Generally I kept most of the core logic **the same**. The main differences are in the embeddings cache - I use a simple synchronous in-memory cache which (currently) does not persist past the session close, though hooking that up to `indexeddb_load` and `indexeddb_save` in future can be considered, I wanted to wrap it up clean first. 

The async calls were wrapped with an asyncRunner which uses yield instead, this preserves most of the program flow while allowing it to work in older browsers. I did simplify the `embeddingDB.computeAndGetIfAbscentEmbedding` method into `DBComputeAndLoadEmbedding`, it ditches the redundant functions (only have 2 operations - Load(regen if needed) and Clear), I also combined the codepaths for the Search Term and the existing DB, previously they were duplicated.

Anyway, if you want, you **should** be able to run your version alongside mine, just swap `getRankedEmbeddings` for your class-based one. Alternatively, you can swap at a deeper level too, at the database-proxy functions `DBClearEmbeddings` and `DBComputeAndLoadEmbedding`. Hopefully it's not too chaotic to make sense of. Lastly, you can ditch the final 3 commits alltogether if you prefer - each commit is standalone and in a functional state, so pick and choose which ones you want.

Let me know if any further issues! Otherwise, I plan to merge this within a day.

LostRuins (approved):

henk717:
#UpstreamWhen :P

PR #124, opened 2025-06-14
feat: enhance word search functionality with n-gram parsing to match …

trincadev:
Hi @LostRuins, I checked your refactor and I have here some feedbacks with some changes I would like to propose.

Bad code practices:

- Assigning potentially arbitrary strings to `.innerHtml` is BAD, please avoid it at least on new code (see <https://www.geeksforgeeks.org/javascript/what-is-the-disadvantage-of-using-innerhtml-in-javascript/> and <https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#replacing_the_contents_of_an_element>).
- Using "==" for comparison (like in <https://github.com/LostRuins/lite.koboldai.net/blob/ab22e039f7943ec3902c040fdd05caa4ed5e2bed/index.html#L13266>) is BAD because it causes implicit type coercion, please avoid it at least on new code.

A good practice would be also start adding E2E tests with frameworks like playwright.

Bugs:

- You use type="text" for the input element id="wordsearch_input": this on android devices (not sure on android tablets - I don't have one - but for sure on android phones with google chrome) will remove the "Enter" button from keyboard user phones (see <https://stackoverflow.com/questions/72369349/google-chrome-101-on-android-shows-tab-key-on-so.ft-keyboard-instead-of-enter-for>, I ended using type="search" on the my input search element).
- Switching from read-only to edit mode (and reverse) will break the word indexes. To avoid this you should modify btn_editmode() to re-run also the word search function to recalculate again the word indexes each time users toggle the edit mode.

Functional bad decisions:

- Hardcoded ngram number? Why? You could simply add a number input field setting (I added min=1, max=9, value=2 in my implementation, see <https://github.com/trincadev/lite.koboldai.net/blob/369f7f2a74c15df85faac9b8a5b7ec66899478f8/index.html#L25275>)
- the full text search will match also substrings and this will cause, in my opinion, two wrong behaviours:
  - it put togheter words with very different meaning like "the" and "they" (and remenber that one of the requests was to automatically ban duplicated words... we risk to ban the wrong words)
  - it match also substrings, e.g. the "a" letters (in your initial Harry Potter's txt example - I extracted the raw text for this check - there are 5299 'a' letters, but only 288 single 'a' words). This is also the reason, I think, for which there are some performance issues in your current refactoring with the Harry Potter's example json text.

I hope you will like my proposed changes. Right now I'm keeping the assigns to .innerHtml because I want your feedback for my changes before doing too much work.
Also the regex I'm using:

```
const WORD_REGEX = /[A-Za-zÀ-ÖØ-öø-ÿ0-9"'-’]+/g;
```

It's probably incomplete but I think you have something like this already in the project, I'm right? I'm also trying to see if it's better to do the contrary, matching instead the punctuation signs and splitting/stemming this way.

However, I'm still working on preparing up-to-dated playwright tests to

- finish the refactor:
  - safely remove the assigns to `.innerHtml`
  - decrease the cognitive complexity of `trigger_wordsearch_results()` and `trigger_wordsearch_candidates()`
- after clicking on the results of `trigger_wordsearch_results()`, I want to emphasize ONLY the selected row (I did it in my old implementation)

### DIFF
--- a/index.html
+++ b/index.html
@@ -22782,7 +22782,9 @@ Current version indicated by LITEVER below.
             let count = candidateItems[i][1];
             const btn = document.createElement("div");
             btn.style.border = "1px solid #646464";
-            btn.innerHTML = `<a href="#" id="id-candidate-${i}-nth" aria-label="id-candidate-${i}-nth">
+            btn.id = `id-div-candidate-${i}-nth`
+            btn.setAttribute("aria-label", `id-div-candidate-${i}-nth`)
+            btn.innerHTML = `<a href="#" id="id-a-candidate-${i}-nth" aria-label="id-a-candidate-${i}-nth">
                 <span class="color_wordsearch_surrounding" id="id-key-count-${i}-nth" aria-label="id-key-count-${i}-nth">${key} (${count})</span>
             </a>`;
             let searchstr = key;
@@ -22872,6 +22874,8 @@ Current version indicated by LITEVER below.
         matchRanges.forEach(({ range, preview }, i) => {
             const btn = document.createElement("div");
             btn.style.border = "1px solid #646464";
+            btn.id = `id-div-${idParent}-range-${i}-nth`
+            btn.setAttribute("aria-label", `id-div-${idParent}-range-${i}-nth`)
             let nearbytext = extract_surrounding_text(range, 8);
             btn.innerHTML = `<a href="#" id="id-${idParent}-range-${i}-nth" aria-label="id-${idParent}-range-${i}-nth">
                 <span id="id-${idParent}-wordsearch-result-before-${i}-nth" aria-label="id-${idParent}-wordsearch-result-before-${i}-nth" class="color_wordsearch_surrounding">${nearbytext.before}</span>
@@ -23029,7 +23033,7 @@ Current version indicated by LITEVER below.
 						<option value="1">Alphabetical</option>
 					</select>
 					</div>
-					<div id="wordsearch_results" style="height:calc(100% - 98px);overflow-y: auto;"></div>
+					<div id="wordsearch_results" aria-label="wordsearch_results" style="height:calc(100% - 98px);overflow-y: auto;"></div>
 				</div>
 			</div>
 		</div>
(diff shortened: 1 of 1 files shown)


LostRuins:
oops deleted the wrong branch

LostRuins:
edit

LostRuins:
Tested and it looks good! Keep it up. I am ready to merge this anytime.

LostRuins (approved):
just say whenever you are ready and i will merge

trincadev:
While doing some tests I found a major issue on Safari (both on desktop and mobile versions): the way we scroll the viewport now doensn't work (the viewport scroll only few pixel, but I think the text selection works).
Any idea about how to fix it? I'm trying to debug this but it could take a while, also because tomorrow I have a busy day.
Let me know if have any idea about how to fix it. Also, let me know if you want to pause the pull request until we found a way to fix the safari problem.
Another thing: I didn't add the input numeric field to configure the ngram number because I had css issue when I tried.

Also, please check the last two commits: I did two merge and I hope I didn't made a mess.

LostRuins:
I think you broke a few cases with your new changes. 

**Yours:**
image

New bugs:
- only 1 instance of YOU found instead of the 3 in the document
- Finding weird substrings that do not exist in the document, because you broke my line splitting `at massayou can`
- No longer finding all substrings of search query in first result, like it does previously. First query should match ALL substrings.

**Compare with current main:**
image
- Correct number of 'you' substrings = 3
- No weird substrings results

Here's the story for your testing
[test_story.json](https://github.com/user-attachments/files/20772526/test_story.json)

LostRuins (changes requested):
Please review

LostRuins:
I took a look at the safari scrolling issue, it just required adding 1 line of code to fix the initial scroll position, this should resolve it
https://github.com/LostRuins/lite.koboldai.net/commit/798cba781033e07f2ee43665cb7a7434a5b09430

Also I think that, considering everything else seems to be functioning smoothly already in it's current state, further changes would not be necessary. Thank you for the time and effort you have put in.

Issue #94, opened 2024-11-20
Description on how to contribute to the project


benjaminfoo:
Hello there, 

great project you've created there - I am just looking for some information, so not a "real" issue here: 

How do you develop this page? As the index.html contains 20k of lines, its a lot of code to read when looking into the project. 
Is this just a compilation of multiple files? If it is, are those files available? Or do you really just update the index page? Is there a reason for not segmenting the logic and style into multiple files? 

I'd really like to support the development of the UI on this project - so thank you in advance!

LostRuins:
Actually there are no other files, it really is just one massive html file.
Of course, it was a lot smaller when I started. Many things have been added since.

The reason why I did it this was is for a few reasons, some intentional and some incidental. 
- I wanted maximum portability in any environment. That means not loading external images, CSS or JS resources at runtime, everything the program needs should be bundled together so it can be instantly dropped in and used anywhere with no dependencies.
- I wanted something easy to edit and iterate on. That means using Vanilla JS instead of more customized frameworks - any change can be instantly tweaked directly.
- Adding to the above point, that meant no build process, so no webpack. This keeps class and ID names fixed so mods can work across multiple versions, and simple to modify as anyone with a plain text editor can modify whatever they need, with no build pipeline.

As for how I develop it, I simply use VSCode and add or change whatever I need in this file. It is a little messy and I do want to eventually reorganize it a bit, but most of the internal functions are segmented somewhat logically. You have your CSS classes and styles at the very top, base64 encoded resources for images/sounds/etc, JS constants such as url endpoints, followed by the various features split into individual functions, then finally the html skeleton at the bottom. 

If there's something in particular you wanna add or modify I can help direct you to where it can be found/added/changed.

ehoogeveen-medweb:
Would you be okay with adding some sort of build step if it were sufficiently automated (e.g. triggered on save), or are you totally opposed? Depending on your requirements I could probably whip something up, so I'd like to know if you're interested.

Having the files split up would also make it much easier to add things like linting and code formatting and maybe target older browsers with babel or similar. It's not impossible to add those now (aside from babel), but not without adding a bunch of config files and a `package.json` (which would also be needed for a build step).

ehoogeveen-medweb:
I contributed `split` and `join` scripts in https://github.com/LostRuins/lite.koboldai.net/pull/127.

Those should make it a bit easier to edit the files and process them with tools, but there are some caveats:
[...]

PR #125, opened 2025-06-17
Make indentation more consistent

ehoogeveen-medweb:
A very minor cleanup as I continue to poke at https://github.com/LostRuins/lite.koboldai.net/issues/94

`index.html` uses tabs for indentation, but some spaces have crept in here and there. I went through and turned them all into tabs, trying to match the existing indentation.

The reason this affects conversion is that I'd like to remove the base indentation when writing the individual files - but the indentation can't be detected if it's inconsistent. Unfortunately this does make things a bit fragile, but on the plus side consistent indentation is good, right?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1455,7 +1455,7 @@ Current version indicated by LITEVER below.
 		padding-left: 10px;
 		padding-right: 10px;
 		padding-top: 7px;
-        border: 1px solid #bbbbbbaa;
+		border: 1px solid #bbbbbbaa;
 	}
 	.cht_inp_bg_inner
 	{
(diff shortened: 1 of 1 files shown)


ehoogeveen-medweb:
You can add a `.vscode/settings.json` file and set `"editor.formatOnSave": true` to automatically format on save, although I don't know how much formatting VSCode's built-in formatter actually does. I can check this later.

These options will override user settings in VSCode, so you probably don't want to make them too strict (which might limit how much you can configure).

For more advanced formatting you'd have to choose a dedicated formatter like Prettier, which can be configured via `.prettierrc`. Prettier is opinionated, but if you don't have strong opinions about formatting (or if you agree with its opinions) then it's fine. I think it works even if you don't install Prettier locally via a (development) dependency, as long as you have the extension installed. You can also give an extension recommendation via `.vscode/extensions.json` so anyone who opens the directory in VSCode will be encouraged to install it.

There are alternative formatters though, like `Biome` and `dprint`, so if you're looking beyond VSCode's built-in formatter then it would make sense to explore the space.

PR #127, opened 2025-06-20
Add scripts to split and join the main html for easier editing

ehoogeveen-medweb:
This turned out to be much more complicated than I was hoping for; I apologize for the amount of code and I hope it's not a deal breaker!

The first problem is that all the HTML parsers I could find treat whitespace as something to normalize or ignore, especially whitespace between attributes. This makes true two-way conversion impossible. I eventually found [`html-lexer`](https://www.npmjs.com/package/html-lexer), which explicitly preserves whitespace. The downside is that it only produces a stream of tokens, so I had to write a rudimentary parser and serializer to actually use it!

The second problem is that the embedded JavaScript and CSS have some leading and trailing newlines, and are indented to match their start tags. This makes for ugly files, and the additional indentation would get removed by any code formatter, which would break the conversion. To work around this, I added a simple algorithm to remove leading and trailing newlines, and remove the base level of indentation.

All in all it's a fair amount of code; I tried to add enough comments to make it understandable. I also used JSDoc comments to document the types; typescript would be better, but the current Node.js LTS (22) doesn't quite support that yet without passing the `--experimental-strip-types` flag (maybe soon though).

The `split` script added by this commit works as follows:
1. Parse the input HTML file into a custom parse tree
2. For every `<script>` element with contents:
    1. Get the target filename for the `id` attribute or generate one
    2. Remove extra leading and trailing newlines and indentation
    3. Record the details of the changes in a `data-actions` attribute
    4. Write the embedded JS to the target file
    5. Reference the outlined JS through the `src` attribute
3. For every `<style>` element with contents:
    1. Get the target filename for the `id` attribute or generate one
    2. Remove extra leading and trailing newlines and indentation
    3. Record the details of the changes in a `data-actions` attribute
    4. Write the embedded CSS to the target file
    5. Turn the `<style>` element into a `<link rel="stylesheet">` element
    6. Reference the outlined CSS through the `href` attribute
4. Serialize the updated parse tree and write the HTML file

The `join` script reverses the process, applying the `data-actions` and inlining the JS and CSS (removing the previously added attributes and turning `<link rel="stylesheet">` back into `<style>`).

To run them:
1. Run `npm install` to install the `html-lexer` dependency (which itself is dependency free)
2. Run `npm run split` or `npm run split index.html` to generate the split files from the monolithic HTML
3. The split files are placed in directory `split` (subdirectory `index` for `index.html`) and are untracked
4. Run `npm run join` or `npm run join index.html` to turn the split files back into the monolithic HTML, replacing the original file

It also works for `koboldcpp_api.html` and `kudorank.html` by passing them as the argument instead, although their indentation is a bit inconsistent so the generated `.js` and `.css` files don't look great.

You can run the generated HTML files in your browser just like the original - all the references work. As a simple follow-up we can give each element an `id` attribute to improve the output filenames and tidy up the indentation a bit more, then maybe configure some formatting and/or linting.


henk717:
kcppapi and kudorank are actually seperate things hosted on the same domain

ehoogeveen-medweb:
Well, I'm just saying the script technically works on them - the output is untracked either way :)

henk717:
Gave it a try but it differs quite a lot from what I imagined would be useful. My original idea was splitting based on markers in the original file so you can easily see what to open. This split script instead seems to split into arbitrary js names which to me personally would just make it more complicated rather than easier but i am not the target audience. Whats the value for you compared to the single file?

ehoogeveen-medweb:
More meaningful filenames can be created by assigning `id` attributes to each `<script>` element. You could also split them up more by splitting the code across more `<script>` tags (which is just a matter of closing one and opening another).

For this PR I focused on getting something that works unconditionally, but the experience can be improved without major changes.

ehoogeveen-medweb:
I pushed a commit that adds ids to the existing `<script>` and `<style>` elements, so at least they don't get generated names anymore. But the "main" script block is still huge and I'm not sure where to split it up - I tried feeding it into several LLMs and they just got lost trying to analyze it (they'd do well at the start and at the end but skip over a huge chunk in the middle).

Hopefully this commit at least shows the basic idea, though.

LostRuins (approved):
Looks good, seems to have worked. 

image

Thank you for your contribution

LostRuins:
Alright I added a bare bones readme, but feel free to update it if you wish.

PR #128, opened 2025-06-21
Add formatting using dprint

ehoogeveen-medweb:
This uses [`dprint`](https://dprint.dev/) to format the code and adds some VSCode settings to help work with it. The advantage of dprint over prettier is that it's less opinionated and more configurable, and its line width is an actual (soft) limit and not an aspirational value. The disadvantage is that the standard VSCode extension doesn't *bundle* dprint, so you have to install it globally or run `npm install` to get it set up. But it's a one time step and then it works.

I tried to choose some reasonable defaults, although I wasn't sure what to do about the brace style. A bunch of places use this style:
```js
if (condition)
{
	// do something
}
```
But a lot of places also use this style:
```js
if (condition) {
	// do something
}
```
The 1st style is mostly applied for `if` statements - functions, for loops etc. generally use the 2nd style. In the end I decided to go with the 2nd style for consistency, but let me know if the 1st style should be used instead.

The line length configuration is a bit strange - it seems like tabs are always counted as 8 spaces, and so to get the HTML formatting to agree with the JS and CSS formatting, the HTML line length needs to be 2x8 spaces more (166 instead of 150). With this configuration, you can format the monolithic HTML or the split files and get the same result.

Actually seeing the change in this pull request is a bit hard, as GitHub just gives up trying to render the diff :\ VSCode also struggled until I set its diff algorithm to "legacy".

Instead, I made a 2nd pull request on my fork where I started tracking the split files to make them visible: https://github.com/ehoogeveen-medweb/lite.koboldai.net/pull/2. You can expand the individual files there, although the changes to `main.js` are still too big for GitHub to display them.

Alternatively, you can check out this branch and look at the commits in VSCode, or take the first few commits and play with the formatting locally.

Let me know what you think!


LostRuins:
Would have been nice to have set this up long ago at the start of the project, but I think applying it full scale now would probably be quite disruptive on the git history and on other downstream projects like esolite/esobold. 

I think perhaps we should stick to manual linting for now - if you notice any particularly bad spots let me know.

ehoogeveen-medweb:
Hmm, it appears that the VSCode extension for dprint doesn't support formatting selections within files. Given that a full reformatting seems like a no-go, that might be a deal breaker for dprint overall. I'll look into the other options some more.

LostRuins:
Yeah I think we can put this on hold for now

ehoogeveen-medweb:
I hope to come back to this at some point, but probably using Prettier (unless the landscape changes significantly). Closing for now.

PR #130, opened 2025-07-27
feat: Agent mode

esolithe:
Adds an agent logic loop to the instruct mode of KLite.

Features:
- Agent mode logic loop, configuration options, along with much of lites functionality automated through it as steps

At this point in time this PR has several limitations:
- KCPP API specific
- Does not include the latest audio -> LLM stuff, but does support whisper / llava
- Highly untested (it's been tested for months in Esobold, but I mean in terms of the cherry pick / merge) given the scope of the code being pulled across 
- Excluded additional libraries (such as math.js)
- Is written using async / await rather than generators
- Does not have model switching configured for the KCPP variant of it (when compared to Esobold - configs probably work fine, but I don't think current model / current confit requests exist in KCPP.  These will need to be migrated across)

As you can see, fairly long list of stuff left above which needs to be done to poet it across.  At this moment, this is about as far as I have got - extracting the code and such.

Comments / thoughts appreciated!

### DIFF
--- a/index.html
+++ b/index.html
@@ -28754,12 +28754,6 @@ let checkFinalThoughtsPrompt = `Action: {"command":{"name":"thought","args":{"me
 		document.getElementById("agentCOTMaxnumeric").value = localsettings.agentCOTMax;
 		document.getElementById("agentCOTRepeatsMax").value = localsettings.agentCOTRepeatsMax;
 		document.getElementById("agentCOTRepeatsMaxnumeric").value = localsettings.agentCOTRepeatsMax;
-		document.getElementById("disableSaveCompressionLocally").checked = localsettings.disableSaveCompressionLocally;
-		document.getElementById("enableRunningMemory").checked = localsettings.enableRunningMemory;
-		document.getElementById("enableAutosaveToServer").checked = localsettings.enableAutosaveToServer;
-		document.getElementById("worldTreePrune").checked = localsettings.worldTreePrune;
-		document.getElementById("worldTreeDepth").value = localsettings.worldTreeDepth;
-		document.getElementById("worldTreeShowAll").checked = localsettings.worldTreeShowAll;
 	}
 
 	confirm_settings = () => {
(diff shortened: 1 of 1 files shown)


LostRuins:
Thanks for the PR.

tbh i know @henk717 wants this upstreamed, but in its current state i'm afraid this PR is way too large for me to do anything with. 
For me I think i'll probably find it easier if I break everything down into a few steps:

Phase 1:
- First, port the agentic tool calling API fields from the OAI compatibility endpoint into kai endpoint, since tool calling API already works for third party agents. Known to work on **semantic kernel** and **autogen**. This is purely a backend thing and doesn't involve lite.
- Then, setup the main agent loop that can 
  - populate the enabled tool schemas for each request, 
  - evaluate which function calls to call and their args, 
  - then execute the functions with provided arguments and return the results into context
- At this point we would have something like https://github.com/ggml-org/llama.cpp/pull/13501
- Since Lite's context lacks the concept of "turns" and permits freeform text, some scaffolding must be added to keep track of current tool and tool results. I would think that this should be part of context - the agent itself is mostly stateless.

Phase 2:
- Then once done, look through each of your tools "**executor**" blocks for tools and port the desired ones in two parts:
  - Tool call schema (description and args) in [OAI format](https://platform.openai.com/docs/guides/function-calling?api-mode=chat)
  - Tool call functionality (execution) as a JS function
- You have 18 total tools now and I'm not sure all of them will be a good fit. 
- Allow user to toggle enabled/disabled each tool, or add their own tools eventually.
 
Check out the agent frameworks i've been messing around with that work in koboldcpp. I used **semantic kernel** and **autogen** which follow the openai tool calling protocol. 

These currently work in latest experimental mainline kcpp.

This one allows me to execute SQL queries to read from a sqlite database.
[autogen_agent.py.txt](https://github.com/user-attachments/files/21508639/autogen_agent.py.txt)

This one demonstrates multi step agentic tool calling -> Get menu -> get soup item -> get soup price.
[semantic_agent.py.txt](https://github.com/user-attachments/files/21508640/semantic_agent.py.txt)

I think this approach would match openai's expectations of tool calls, rather than doing something too different. 
- Moreover, this is a massive undertaking to add. Huge, messy feature that will likely take a lot of time to add and I'm kinda busy nowadays so I dunno when I can get to it too.
- It is not entirely congruent with the work that you've already done, and I don't want to disrupt your own efforts in esobold, just because we have different design decisions.
- You have additional flows that this agentic tool calling approach does not support - mainly things like clicking images in the ui to pick them, or selecting an option from a few AI generated multiple choices.
- So i think the best approach would be to add tool calling capabilities as a backend feature first (without changing lite), then add the rest of the stuff in phase 1 in a way that is non disruptive to your existing implementation.
- then once that's done, you can see which (if any) esobold tools you want to migrate to this approach, while I explore an extensible way for new tools to be added.

henk717:
```
You have additional flows that this agentic tool calling approach does not support - mainly things like clicking images in the ui to pick them, or selecting an option from a few AI generated multiple choices.
```
I don't see why not, those choices are a tool call the LLM invokes.

esolithe:
Closed as tool calling in a different form has been added to Kobold - agent mode conflicts with the approach, so it doesn't make sense to keep this open.

New MCP UI looks neat though!

PR #138, opened 2025-11-22
Added Lora Support through ComfyUI API

Yasakikondo74:
Screenshot 2025-11-22 202710


Yasakikondo74:
Tbh, I probably made errors somewhere, I only do some edits to get lora working with comfyUI and it is working right now

PR #142, opened 2025-12-21
More careful Tavern imports and prompt construction

hu-yijie:
This PR contains a bunch of tweaks to handle some edge-cases that I've run up against in Lite while messing around with (third-party) Tavern cards, and which I figure would be worth upstreaming. (Since each of these changes are relatively small, I've put them all together.)

---

1. **Alias-handling causes `undefined` to appear in the memory**

Fixes a bug occuring in `function load_temp_scenario_from_tavernobj(obj)` when `examplemsg==""` and the alias `obj.mesExample` is undefined.

2. **"Smarter" memory truncation that prioritizes pruning example messages**

When the memory gets too large for the context, Lite's current behavior is to truncate from the start (using `substring_to_boundary()` or the `substring` method). However, the [V1 character card spec](https://github.com/malfoyslastname/character-card-spec-v2/blob/main/spec_v1.md) recommends pruning example messages (which Lite currently places at the *end* of memory), and the botmaking community [seems to take for granted that examples will be pruned before the main description](https://rentry.org/NG_Context2RAGs#character-description-advanced-definitions).
Personally, I also agree that pruning `examplemsg` first before touching the main memory makes sense (I don't think I've ever wished otherwise), so I've written a function `truncate_memory()` that opportunistically starts truncating from the first example message (detected by the `"<START>"` string, which should occur at the start of `examplemsg`, but we manually prefix it if not).

3. **Import v2 fields**

The current Tavern card loader discards the [v2 fields](https://github.com/malfoyslastname/character-card-spec-v2/blob/main/spec_v2.md) `creator_notes` and `post_history_instructions`, even though they often contain useful details. I propose to import and put them in the (currently unused) "Personal Notes" and "Author's Note" fields, which seems to match how they're intended to be used (both by the spec, as well as in practice).

(I've also taken the opportunity to change the header for the `obj.description` field to `"Description: "` from `"Persona: "` (Tavern only, other formats unchanged), since there are several Tavern cards where {{char}} is not a character but a setting, and this field really is intended to be a description.)

4. **Handling the `{{original}}` placeholder**

As described in the v2 spec above. Personally I think it's silly to have a placeholder with two different meanings (depending on whether it appears in the `system_prompt` or `post_history_instructions` field), but I have seen cards using it in the wild, so we might as well handle it properly.

From what I understand, it's meant to be replaced with the user's current system prompt or jailbreak (both of which are only activated in Instruct Mode). My proposal is to add the following logic to the `replace_noninstruct_placeholders()` function:
- If the user is not in Instruct Mode, replace `{{original}}` with the empty string `""`;
- If the user is in Instruct Mode, and there is at least one occurrence of `{{original}}` in the memory or A/N (i.e., `system_prompt` or `post_history_instructions`), then we replace it with the Lite system prompt or assistant jailbreak respectively. Hence, when assembling the Instruct prompt for submission, we can skip over the injection of the sysprompt/jailbreak.

---

All these should help to reduce the gap between the out-of-box user experience of chatting with Tavern cards in Lite, and the experience intended by the cardwriter (who is unlikely to be writing with Lite in mind).

### DIFF
--- a/index.html
+++ b/index.html
@@ -5192,7 +5192,7 @@ Current version indicated by LITEVER below.
 		inputtxt = replaceAll(inputtxt,instructsysplaceholder_end.trim(),get_instruct_systag_end(false));
 		return inputtxt;
 	}
-	function replace_noninstruct_placeholders(inputtxt,escape=false,isgametext=true,persistent_countmap=null,original=null)
+	function replace_noninstruct_placeholders(inputtxt,escape=false,isgametext=true,persistent_countmap=null)
 	{
 		let firstopponent = localsettings.chatopponent;
 		if (firstopponent && firstopponent.includes("||$||")) {
@@ -5211,10 +5211,7 @@ Current version indicated by LITEVER below.
 			inputtxt = replaceAll(inputtxt,"{{char}}",(localsettings.chatopponent?firstopponent:defaultchatopponent),true);
 		}
 
-		if(original !== null) //replacing with "" is allowed
-		{
-			inputtxt = replaceAll(inputtxt,"{{original}}",escape?escape_html(original):original,true);
-		}
+		inputtxt = replaceAll(inputtxt,"{{original}}","",true); //redundant in Lite; use sysprompt or assistant jailbreak instead
 
 		for(let i=0;i<localsettings.placeholder_tags_data.length;++i)
 		{
@@ -5303,12 +5300,12 @@ Current version indicated by LITEVER below.
 	}
 	//if alwaysreplace, then settings are not considered, otherwise checks settings
 	//if isgametext is true, then keeping the random values stable will be prioritized
-	function replace_placeholders(inputtxt, escape=false, alwaysreplace=false, isgametext=true, original=null)
+	function replace_placeholders(inputtxt, escape=false, alwaysreplace=false, isgametext=true)
 	{
 		inputtxt = replace_instruct_placeholders(inputtxt);
 		if(alwaysreplace || localsettings.placeholder_tags)
 		{
-			inputtxt = replace_noninstruct_placeholders(inputtxt,escape,isgametext,null,original);
+			inputtxt = replace_noninstruct_placeholders(inputtxt,escape,isgametext,null);
 		}
 		return inputtxt;
 	}
@@ -9340,7 +9337,7 @@ Current version indicated by LITEVER below.
 			let examplemsg = obj.mes_example?obj.mes_example:"";
 			let sysprompt = obj.system_prompt?obj.system_prompt:"";
 			let personalnotes = obj.creator_notes?obj.creator_notes:"";
-			let authorsnote = obj.post_history_instructions?obj.post_history_instructions:"";
+			let posthistoryinstructions = obj.post_history_instructions?obj.post_history_instructions:"";
 			let greeting = selectedgreeting;
 
 			//post process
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -15910,6 +15920,24 @@ <h4><li><a href="${reader.result}" class="color_blueurl" target="_blank" downloa
...
 	}
 
+	//Prioritizes truncating example messages in memory
+	function truncate_memory(current_memory, max_mem_len, to_boundary)
+	{
+		if(current_memory.length <= max_mem_len) {
+			return current_memory;
+		}
+		let basictrunc = to_boundary?substring_to_boundary:((str,len) => str.substring(str.length-len));
+		let startidx = current_memory.indexOf("<START>"); //start of examplemsg
```
LostRuins: I think this is fine but one potential concern is what happens if there are, for whatever reason, multiple `<START>` inside the memory? Should we consider that case?

index.html
```diff
@@ -9589,6 +9594,8 @@ <h4><li><a href="${reader.result}" class="color_blueurl" target="_blank" downloa
 			temp_scenario.chatopponent = chatopponent;
 			temp_scenario.prompt = ("\n{{char}}: "+ greeting);
 			temp_scenario.memory = combinedmem;
+			temp_scenario.personalnotes = creatornotes;
+			temp_scenario.authorsnote = posthistoryinstructions;
```
LostRuins: i think this might be slightly risky though I don't actually understand how `post_history_instructions` is used by card makers. Author's note is injected *very* late into the context, think approx 1 paragraph from the front of the prompt, and moves along with it. So whatever text in here will have a huge influence on output. Would it be safer to append it somewhere nearer the start of memory instead? (help me understand)

LostRuins:
Hi, thanks for the PR! So far looks good, I've added some comments.

hu-yijie:
Matching against `<START>\r?\n` should be fine, I think.

But we can't just simply drop-in `lastIndexOf()` to replace `indexOf()` though, because when there are multiple example messages, this would just erase the last one (each new example has its own `<START>` newline), and then go on to munch tokens from the start of memory until the hard limit of `max_memory_len`. This is what we want to *delay* as much as possible (since info at the very start is often nearly as important as info at the very end), by munching from the example messages whenever we can, and this is why the `truncate_memory()` function logic has a branch
`if(max_mem_len < startidx+7+suffix.length) {`
to separate the two cases: where we must touch the start, and where we do not need to.

Though I can see an opportunity for conservativeness in the second case: if there are multiple example messages (i.e., multiple matches of `<START>\r?\n`), we could munch them from back to front, thus delaying the risk of munching into a false positive until it becomes absolutely necessary. I'm not sure it's worth the additional logical complications, but I'm not strongly against it either.

>Alright then we can keep it, but perhaps we can add it right at the end of Memory rather than into Author's note? It should still have the jailbreaking effect without dumping like 2000+ tokens right besides the user's last input (high risk of derailing)

Wouldn't that break the logic of `truncate_memory()`? Since that function relies on the example messages being at the tail end of memory in order to munch safely. (In other words, we do not want `truncate_memory()` to eat into the `post_history_instructions`.)
I'm not against the idea of the final submitted prompt having it in Memory rather than A/N, but that would require the prompt construction logic to be reworked. At the very least we can't simply append `post_history_instructions` to the memory (directly after `examplemsg`) upon import, it might have to be stored in a separately managed variable or something.

hu-yijie:
Yup, I think this should do it!

LostRuins (approved):

LostRuins:
merged. on most cards this seems to look fine, although I did notice the Seraphina card (ST default card) seemed to contain a bunch of extra info after the <START> tags that would eventually be lost. 

default_Seraphina

index.html
```diff
@@ -15910,6 +15923,25 @@ <h4><li><a href="${reader.result}" class="color_blueurl" target="_blank" downloa
...
+		let examplemsg = current_memory.match(/<START>\r?\n/);
+		if(!examplemsg) {
+			return basictrunc(current_memory, max_mem_len);
+		}
+		let examplemsgstart = examplemsg.index+examplemsg[0].length;
+		let suffix = current_memory.endsWith("\n***")?"\n***":"";
+		if(max_mem_len < examplemsgstart+suffix.length) { //discard all examplemsg
+			return basictrunc(current_memory.substring(0,examplemsg.index)+suffix,max_mem_len);
+		}
+		return current_memory.substring(0,examplemsgstart)+basictrunc(currrent_memory.substring(examplemsgstart),max_mem_len-examplemsgstart);
```
LostRuins: `currrent_memory`

how did this typo work for you? did you test it?

LostRuins: nvm i fixed it

currrent -> current

PR #153, opened 2026-01-29
improved display of tool call results

Rose22:
instead of displaying 2 entries for the tool call results, i just make it emit "calling tool (toolname).." when it's calling it, then replace the message with "got results from tool (toolname)" after it gets the results. it looks neater!

image

this can be further improved in the future by adding a collapsible section that lets you show what the results were. but for now this'll do!


Rose22:
uhoh, found a bug. if it calls multiple tools after another, it keeps displaying the name of the first tool

PR #154, opened 2026-01-29
hide chat name in messenger theme in instruct mode when not injecting chat names

Rose22:
looks a lot nicer, especially since without inject chat names, it defaults to the name "KoboldAI".

the way i did it might not be the best way, i couldn't find where the core of the name logic was. i hope this is okay

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -25742,7 +25742,7 @@ Current version indicated by LITEVER below.
 				curr.msg = render_all_media_html(curr.msg,false,true);
 			}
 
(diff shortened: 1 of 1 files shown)


LostRuins:
Rather than force the user, I've added an independent toggle that allows them to choose whether to show or hide the names.

This will apply for all modes, corpo, aesthetic, messenger, and classic with injected names.

https://github.com/LostRuins/lite.koboldai.net/commit/6d2986f8eed0bf123c643825ee66e0c1b15d40b7

PR #168, opened 2026-04-20
Add padding to chat message body

MinecraftPublisher:
Does what it says on the tin. Sets #chat_msg_body's padding-bottom to 3rem. This allocates a small bit of breathing room for Chat Mode, when using the Aesthetics theme. I chose 3rem because it looked the best after trying 1rem and 5rem.

I can also implement this for the Corpo and Normal interface themes as well, in this very same PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -761,7 +761,7 @@ Current version indicated by LITEVER below.
 	}
 
 	#chat_msg_body {
-		padding-bottom: 3rem;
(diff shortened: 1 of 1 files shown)


LostRuins:
i can add this padding, but 3rem is too much, probably 1.5rem is better. I'll change it to that, if you don't mind.

PR #170, opened 2026-04-25
Add automatic server mod loading

MinecraftPublisher:
Right after applying the userMod, tries to pull `./serverMod.js` from the server. If it exists, it is evaluated. If not, this is a no-op.

The goal for this is for a hoster / provider to make modifications to the client without touching the actual code and creating conflicts when updating their copy of KoboldAI Lite.

### DIFF
--- a/index.html
+++ b/index.html
@@ -760,10 +760,6 @@ Current version indicated by LITEVER below.
 		}
 	}
 
-	#chat_msg_body {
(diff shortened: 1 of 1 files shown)


LostRuins:
as discussed in discord, probably not a good idea for loading external JS for now. I think modifying the html if you really need it is still better.

PR #171, opened 2026-04-25
Implement functions to modify dependent var values in a unified way

MinecraftPublisher:
This avoids a usermod having to write seven lines of variable assignments just to change the aihorde base URL.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4058,23 +4058,33 @@ Current version indicated by LITEVER below.
 	var poll_interval_chunked_tts = 300;
 	var max_branches = 6;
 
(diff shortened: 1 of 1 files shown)


LostRuins:
those values are not intended to be changed at runtime actually, they were consts. I only swapped them so usermods have the flexibility to access and modify them.

MinecraftPublisher:
Yes, and this makes it easier for a usermod to modify them.

LostRuins:
Yes but you can place these function directly within your usermod too. 

I don't want to add extra functions that aren't useful within the main UIs code, it adds a lot of unnecessary clutter.

PR #181, opened 2026-05-07
Webmanifest improvements

### LINKED ISSUE #178: Use .webmanifest format
Currently the improper though accepted manifest.json format is used
instead, .webmanifest format should be used per https://w3c.github.io/manifest/#media-type-registration
This change adds a VERY marginal security benefit due to the specific purpose .webmanifest being more limited than .json 
That is, this change moves us from the type application/json to the subtype proper of application/manifest+json

Though the main focus and benefit is improved compliance with the actual specification. To fix this, it is very simple.
Just change the . extension to .webmanifest from manifest.json so you could have manifest.webmanifest or if feeling fancy rename it to something like app.webmanifest
Then change the corresponding invocation of it in index.html and the other places it is accessed, most of the html files.

thegreatGreenstar:
Resolves #178
This makes us properly comply with the webmanifest specification such that we use a .webmanifest
And since the webmanifest is specifically for the website version unlike the favicon stuff in index.html, we may use a reference to the included image here, as the webmanifest should ONLY be used in a PWA context (not discovered, discovered and saved in web context but not used) which per our file structure should be valid.


thegreatGreenstar:
As a note, your behavior for accessing it differs on what should be the same / level between just outright access and ./ access
Both seem valid, though for consistency I recommend choosing one.
It is not the focus of this issue though, so remains unfixed by me here.

thegreatGreenstar:
If you prefer strict linking I can use the full url which would also work for the issue closed about fixing base64 usage.

thegreatGreenstar:
For more information about .webmanifest
https://w3c.github.io/manifest/#media-type-registration

LostRuins:
What's the benefit to this change? Seems like the .json one works well and has better backwards compatibility with older browsers. Does it not work for you?

thegreatGreenstar:
As detailed in https://web.dev/learn/pwa/web-app-manifest the preferred is .webmanifest
We support .webmanifest and should use such. As part of our .webmanifest improvements we should also include more icons, ideally an SVG as well, to scale to any size. (For example a basic vectorized version attached in this comment)
This is a point of discussion in https://github.com/w3c/manifest/issues/689

thegreatGreenstar:
niko

LostRuins:
@henk717 thoughts? I don't fancy losing backwards compat and having additional non-embedded icons for no reason.

henk717:
Personally I don't know of anything that doesnt work in the current setup and I dont think changing for the sake of changing is worth it. We have no reports of the current manifest breaking. Why risk it?

thegreatGreenstar:
How about this
Since the spec for webmanifest is yet to actually be finalized proper, I close this issue as unplanned for now, open a new pull request to change the webmanifest to reuse the already existing favicon niko.png, and maybe use the screenshots (https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/screenshots)
Then when the webmanifest spec is finalized I reopen a pull request to move to .webmanifest if still relevant and we discuss that fully?

PR #187, opened 2026-05-18
Manifest image handling improvements

thegreatGreenstar:
What this does is:
* Allows the PWA to take advantage of [screenshot functionality](https://w3c.github.io/manifest-app-info/#screenshots-member) using our previews.
* Makes the PWA properly reuse niko.png, saving a marginal amount of data, making the manifest much more readable, and just being a generally nice thing.
* Make background color to what ``outerbodybg`` actually is. This provides better visual transition between opening/loading and the actual application as the color doesn't suddenly change.
* Moves sizes to a new line for formatting consistency.


thegreatGreenstar:
Screenshots isn't widely implemented yet, but tends to show on application installation, mainly on android phones.
Though our resolution is insufficient here for most devices, so it won't display on as many devices. It should be enough for some devices at least though.
Images that aren't usable will not be loaded regardless, so this should be a net improvement still.

thegreatGreenstar:
As for using niko.png instead of inline
We have no reason to not do so. favicon I guess we may inline, though strange to do so still
but web manifests are only usable in web and PWA contexts, in which these images are accessible, and thus said images should be used.
once more, this gives us slight benefit with zero downside.

LostRuins:
fine i guess

PR #194, opened 2026-06-28
Add groupchat participant management

### LINKED ISSUE #192: Feature Request: Enable Group Character Selection Checkboxes in Instruct Mode When "Inject Chat Names" Is Active
When using Instruct Mode with "Inject Chat Names" enabled, the group character selection checkboxes that are available in Chat Mode are disabled. Since "Inject Chat Names" already creates a hybrid chat/instruct experience with multiple named characters, it would make sense to enable those checkboxes in this context too.

Right now, switching which character responds next requires going into settings and editing the AI Names list manually. If the checkboxes were available whenever "Inject Chat Names" is active, you could change the next responder inline without ever leaving the conversation, just as if using chat mode directly.

hu-yijie:
**DISCLAIMER**: Yes, this PR was vibe-coded. It shouldn't break any existing workflows though (except possibly the patch in commit 5 below, which I consider a bugfix), and I'm willing to take responsibility for justifying each step of Codex's code.

This started off as an idea for fixing #192 by unifying the group chat interface in Chat and Instruct modes, but then I realized that it also fills the long-standing request for SillyTavern-style multi-character chats, where each participant has their own avatar and injectable character card (called "character contexts" in the code)...

The first commit *0110186* resolves #192, so you can cherry-pick that one even if you disagree with the rest of the implementation.

---
AI-generated summary of the individual commits:

1. *0110186 Unify groupchat UI for Chat+Instruct Mode*: Replaced the split Chat/Instruct selector behavior with a unified
  table-based participant UI. The first row represents the user, subsequent rows represent AI participants derived from
  `chatopponent`, and the Speak/Reply controls now work consistently across Chat and Instruct modes.

2. *9c1a501 Replace `groupchat_removal` with `groupchat_participants` interface*: Removed the session-only `groupchat_removals` model and replaced it with a runtime `groupchat_participants` interface. Reply eligibility is now derived from participant rows and their `selected` state, preserving existing behavior while making future participant metadata possible.

3. *5409e54 Add `localsettings.groupchat_data` for storing extended fields*: Added `localsettings.groupchat_data` as row-aligned persistent storage for participant metadata. `groupchat_participants` now combines existing fields like user/bot names with each row’s stored metadata, while keeping `chatopponent` as the backward-compatible source of bot names.

4. *21cde29 Add backend support for character portraits and context*: Extended backend rendering and prompt substitution to support participant-level portraits, colors, and `char_context`. Participant turns can now render using per-character aesthetics, and `{{Name_context}}` placeholders resolve from the corresponding participant metadata.

5. *f00a3f6 Support char portraits+context frontend, patch repack_postprocess_turn behavior*: Added frontend controls for participant avatar and context editing from the groupchat selector. This also patched `repack_postprocess_turn()` so display-time placeholder substitution respects `localsettings.placeholder_tags`.

6. *70aecca Cleanup groupchat frontend*: Refined the groupchat selector into a central management UI. Participant names became inline editable controls, the selector can directly add, rename, and delete AI participants while preserving row-aligned `groupchat_data`, and the UI styling/copy was cleaned up for discoverability and consistency.

7. *f92ee8c Let Add AI Participant initialize portrait+context*: Extended the direct `Add AI Participant` dialog so new AI participants can be initialized with an optional portrait and character context at creation time. The existing Chat Settings add flow remains name-only, while the groupchat UI constructor passes optional rowdata into `localsettings.groupchat_data`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4393,6 +4393,7 @@ Current version indicated by LITEVER below.
 		chat_context_mod: true, //extra injection for chat mode
 		chatname: "User", //name to use in chat
 		chatopponent: defaultchatopponent,
+		groupchat_data: [], //persistent metadata for groupchat selector rows
 		instruct_starttag: "{{[INPUT]}}",
 		instruct_endtag: "{{[OUTPUT]}}",
 		instruct_systag: "{{[SYSTEM]}}",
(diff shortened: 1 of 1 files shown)


LostRuins:
btw let me know once its ready for review. lets try to keep it not too complicated for a first inclusion

hu-yijie:
It should be ready now! The biggest change that I wanted to make was to replace my original approach of giving each character a `{{Name_context}}` placeholder (whose usage was honestly unclear) with a common `{{context}}` placeholder that resolves differently according to who the current active speaker is (which is more aligned with [how SillyTavern understands it](https://docs.sillytavern.app/usage/core-concepts/groupchats/#group-generation-handling-mode)).
Keeping track of the `active_speaker` turned out to be trickier than expected; I ended up threading it through the `replace_*` functions, similarly to `persistent_countmap`.

That was commit `0a35e7e`; the others are small cleanup/bugfixes.

LostRuins:
I like the core concept but I will probably simplify the code, 800 lines is overkill, and I don't actually think we need the `{{context}}` placeholder. The chat participant meta can simply be attached along with the memory

I will do a bit of refactoring first

PR #196, opened 2026-07-05
tts ai embed fix

mosujiba:
Generated TTS audio will behave like generated images and no longer automatically be sent as multimodal input to the AI.


LostRuins:
Thank you

PR #201, opened 2026-07-19
Added ANIMA support through comfyUI

Yasakikondo74:


Yasakikondo74:
idk if adding clip text encode + vae loader for anima being a bit too much
but anyway @LostRuins please review :3

LostRuins:
I'm not really sure what's all this for.  It should already be working with comfyui in its current state...

Yasakikondo74:
Yeah but that only work for checkpoint models where SDXL like NoobAI, Illustrious or Pony type checkpoints. This here is different where it could work with ANIMA models and potentially flux, Z-turbo models, using diffusion model loaders instead of all in one checkpoint model loader. Which can be sometime lighter and faster or overall better quality output.

PR #202, opened 2026-07-23
Import groupchat participants from character card

hu-yijie:
As mentioned in the Discord, this upstreams the groupchat UI extension to allow importing Tavern character cards. I didn't want to hold on to it for too long, in case you're waiting on it before the next version update.

I've also taken the opportunity to patch a small issue with the `impersonate_message(` function, which currently force-closes the groupchat UI even if you decide to cancel it (by leaving the field blank). While this makes sense for one-on-one conversations, I think it's better for groupchats to keep the popup open. Not a major issue, though.


LostRuins (approved):
looks fine

PR #205, opened 2026-08-06
fix: chat messages breaks when when matched characters in reasoning content

jasonts:
in chat mode, when message like this

```
User: *user's emotion* test.
KoboldAI: *KoboldAI's emotion* response.
User: <think>prelude
KoboldAI: his status
User: user's status
next possible steps</think>user's action
```

since there is a character message prefix (KoboldAI:) in User's reasoning content
the chat message list will be broken.
it even truncate messages away in some cases.

this fix will preserve whole reasoning content without any tamper.

btw, it is fixed by Codex since I did not familiar with frontend,
if you are not comfortable with this,
you can fix the issue by your style and reject this PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6063,6 +6063,54 @@ Current version indicated by LITEVER below.
 		const safeStop = escapeRegExp(localsettings.stop_thinking_tag);
 		return `${safeStart}(${get_thinking_content_regex()})${safeStop}`;
 	}
+	function protect_complete_thinking_blocks(input)
+	{
+		let protected_blocks = [];
+		if(!input || !localsettings.start_thinking_tag || !localsettings.stop_thinking_tag)
+		{
+			return {text:input, blocks:protected_blocks, placeholder_prefix:""};
+		}
+
+		let tag_pairs = [[localsettings.start_thinking_tag, localsettings.stop_thinking_tag]];
+		let escaped_start = escape_html(localsettings.start_thinking_tag);
+		let escaped_stop = escape_html(localsettings.stop_thinking_tag);
+		if(escaped_start!=localsettings.start_thinking_tag || escaped_stop!=localsettings.stop_thinking_tag)
+		{
+			tag_pairs.push([escaped_start, escaped_stop]);
+		}
+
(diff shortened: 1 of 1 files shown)


LostRuins:
I think that in instruct mode, this is already handled, when I send a multi turn message with mismatched thinking tags they seem like they are wrapped correctly?

image
image

jasonts:
Oh, sorry for my bad chat example,
I have write incorrect role between user and AI, and the first 2 lines may also confusing the describe.

i'm using chat mode, not instruct mode, it is parsed correctly when the tag is in the middle.
but the problem is when <think> is the model's first response like this

```
User: send testing 123
AI: <think>
thinking 456
User: thinking 789
AI: thinking 10 11 12
13 14 15 </think> Reply 16 17 18
```

the actual reasoning context is from "thinking 456" to "13 14 15",
it would render like this incorrectly:

圖片

not like this below

圖片

LostRuins:
Are you in chat mode or instruct mode? In chat mode if the AI response has the user's name tag it would indeed become a new turn, as expected.

jasonts:
Chat mode.

Because in my opinion, when the name tag is in normal content, it would become a new turn.
But when the tag is appear in reasoning_content, it would treated as part of the reasoning content, not to a new turn.

So do I need to prompt it, to let AI not to output name tag in reasoning block?

LostRuins:
Well if the name tag is actually in the reasoning block it is fine. The problem is:

1. The AI output name tags in reasoning block, and
2. The reasoning block is incomplete due to your output length being too short or too much thinking without limiting it
3. The mismatched block causes the name tag to leak into the normal text and turn into a new section.

jasonts:
i see. the problem seems we cannot control the length of the reasoning content.
since I am use llama.cpp as LLM backend,
and llama-server has POST /v1/chat/completions/control that can stop reasoning immediately, and start output normal content.

But, that api did not works on all models though. Gemma 4 31B works, Muse Glimmer isn't.
also didn't know other API provider has similar API or not.

LostRuins:
Actually in koboldcpp you can control it.  open koboldai lite, In the Settings > Tokens page you can set a reasoning effort budget

PR #206, opened 2026-08-07
[Feature] Custom Javascript tools + slash commands

hu-yijie:
This PR allows users to define custom Javascript functions that hook into Lite's tool calling mechanism (i.e., they present as tools to the AI, without the need for an MCP server).
In addition, the user may enable individual custom tools to be callable using /slash command syntax (this is a separate setting from whether the tool is enabled for the AI):
custom_tools
*(Edit to add: I just noticed that the image above is outdated -- the placement of the "Custom Tools" button has been moved into `Settings > Tools > MCP Tool Calling`, in order to not to be too discoverable for inexperienced users.)*
The main use case I have in mind is to emulate the agentic experience of e.g. Codex, but with full control of the AI's autonomy (since we have complete oversight of the catalogue of command names and their execution effects).
In particular, since the command vocabulary would be tied to the context of the user's session, I'm saving the custom Javascript code as an additional setting in the user's `.kcpps` file. To mitigate security risks, none of this code is evaluated until a custom tool is actually executed (i.e., "just-in-time") either by the user or the AI, and every bit of code that gets loaded is fully visible and inspectable from within the custom tool manager UI.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4643,7 +4643,9 @@ Current version indicated by LITEVER below.
 		saved_comfy_bearer_token: true,
 		saved_xtts_url: true,
 		generate_images_mode: true,
-		saved_mcp_urls: true
+		saved_mcp_urls: true,
+		custom_tools: true,
+		custom_tools_AI_enabled: true
 	};
 
 	const defaultsettings = JSON.parse(JSON.stringify(localsettings));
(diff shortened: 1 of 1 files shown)


LostRuins:
cool, this might be what @henk717 and @esolithe wanted for lite native tool calls. The change is huge tho, so review might take some time

LostRuins:
Codex strongly suggested sanitizing the custom tools instead of running it through a raw eval so I let it change that part but i guess its functional enough so we can probably merge.

LostRuins:
Don't really like the slash command thing since it just writes directly into the input box, but i'll leave it in

hu-yijie:
> Don't really like the slash command thing since it just writes directly into the input box, but i'll leave it in

Yeah, I've also considered pushing the return value as a user turn, which would allow the AI to see that the user made a tool call. But I eventually decided against it, partly because this could confuse the model, and partly because the returned tool call result itself doesn't actually get displayed (the user can see it by switching to Edit Mode, but we can't expect them to realize this).

In any case, I'm fine with the current location, but I wouldn't object if some other user suggests a better place to move it. The same goes for the slash command syntax; I have no qualms with other people PRing tweaks to the parser in order to better handle edge cases.

I've tested the latest version (9828113) and can confirm that it does handle the simple example tools correctly. However, for tools that return base64 media data, it does seem to trip up multimodal models:

Issue:
1. Load and enable a tool that returns an image (I've attached a test example, which generates an image from an A111 endpoint).
[image_generation_tool_test.json](https://github.com/user-attachments/files/31350221/image_generation_tool_test.json)
2. Prompt the AI so that it calls the tool (e.g. "Can you show me what you look like?" should suffice)
3. Expected: The returned image gets CLIP-encoded for the next toolcall reasoning stage. Context remains clean.
Actual: The base64 string is left in context (and is usually long enough to get truncated and/or confuse the model).

The crux of the problem seems to be the vision formatting being messed up by the stringification on line 9360:
```
callresp.content = JSON.stringify(outitems);
```
Removing the `JSON.stringify(` call (i.e. leaving `callresp.content = outitems;` as an array) results in the Expected behavior. The other simple example tools also continue to work as intended.

Is the stringification here really necessary, given that the very next step is to stringify the entire `callresp` object anyway (line 9362)? If it is, maybe we could carve out the base64 media exception here, just before the stringification.

hu-yijie:
Also,  on the lite-beta mobile UI, I spotted a couple of small issues with the custom tool editor:

- (Line 33884, `id="customtooleditor_name"`): For the tool name input textbox, `style="width: calc(100% - 250px);"` shrinks it too much; 100px should probably suffice.
- (Line 33889, `id="customtooleditor_parameterstemplatecontainer"`): The parameters table is spilling out of its container; I believe this can be fixed by adding the attribute `class="settinglabel"`.

LostRuins:
Should be fixed in https://lite-beta.koboldai.net, could you check if its ok now?

hu-yijie:
Yup, I can confirm that both issues are resolved now. Thanks!
(If I had to nitpick, I feel like the "Name" column is a bit too narrow; maybe changing the last two columns to something like "Default Val." (content: "Set?") and "Del." could alleviate this. But in any case, I'm not actively using the mobile UI...)

I've also tried out the `dev` branch and can confirm that it successfully resolves the base64 media issue.

PR #214, opened 2026-09-20
Fix(?): Allow customtools to be imported from savedsettings

hu-yijie:
Noticed this behavior some time ago, but I've only recently had the time to look into this...

Although custom tools created or loaded by the user are stored in `localsettings` and hence exported to their KAI savefiles, because these fields are currently treated as protected, these tools are silently lost when the KAI file is re-imported. I doubt this is the intended behavior, so I've tweaked the `kai_json_load` function to restore customtool import, while adding a checkbox to the Advanced Load options so that the original behavior is still possible.


LostRuins:
No, the custom tools are not supposed to be exported to savefiles. Because they can run arbitrary javascript, they work like usermods, they are intended to be set and used for the local user only. Otherwise, it's easy to create malicious savefiles containing dangerous tools.

In fact, generating a json savefile redacts the custom tools entirely 
`new_save_storyobj.savedsettings.custom_tools = [];`

hu-yijie:
Ah, I *did* have a few files with custom tools saved in them, but after checking more carefully I noticed that they were all generated when I was still experimenting with the beta versions of this feature.

I agree that these tools are basically usermods, so the policy does make sense. For local usage, it's not a serious impediment, since the custom tools array can still be exported and re-imported as a separate JSON file.

